PR #5, opened 2019-11-27
ADCM-999 check hostcomponent when host is removed from cluster

grey-kristy:


Maunty:
Bender build

Maunty:
Bender build

PR #55, opened 2019-12-16
ADCM-1027 Refactoring transaction.atomic for task and job runners

a-alferov:


cm/job.py
```diff
@@ -657,30 +654,34 @@ def restore_hc(task, action, status):
...
+    api.save_hc(cluster, host_comp_list)
 
 
-@transaction.atomic
 def finish_task(task, job, status):
-    action, obj = set_action_state(task, job, status)
-    unlock_objects(obj)
-    restore_hc(task, action, status)
-    set_task_status(task, status)
+    action = Action.objects.get(id=task.action_id)
```
grey-kristy: We can not find this Action and fall with exception. Try/except is need here

PR #59, opened 2019-12-19
ADCM-645 rewrite adcm check logic to serialize all checks of one job via DB

grey-kristy:


cm/job.py
```diff
@@ -729,20 +731,22 @@ def log_check(job_id, title, res, msg):
...
         err('JOB_NOT_FOUND', f'no job with id #{job_id}')
+    cl = CheckLog(job_id=job.id, title=title, message=msg, result=res)
+    cl.save()
+    return cl
 
+
+def finish_check(job_id):
+    data = []
+    for cl in CheckLog.objects.filter(job_id=int(job_id)):
+        data.append({'title': cl.title, 'message': cl.message, 'result': cl.result})
```
a-alferov: We can use the inline method values() for QuerySet object.
```
data = list(CheckLog.objects.filter(job_id=int(job_id)).values('title', 'message', 'result'))
```

cm/job.py
```diff
@@ -729,20 +731,22 @@ def log_check(job_id, title, res, msg):
             err('JOB_NOT_FOUND', f'job #{job.id} has status "{job.status}", not "running"')
     except JobLog.DoesNotExist:
         err('JOB_NOT_FOUND', f'no job with id #{job_id}')
+    cl = CheckLog(job_id=job.id, title=title, message=msg, result=res)
+    cl.save()
+    return cl
```
a-alferov: Is this `return cl` necessary here?

cm/job.py
```diff
@@ -729,20 +731,22 @@ def log_check(job_id, title, res, msg):
...
-        raw = f.read()
-        data = json.loads(raw)
-        f.seek(0)
-    else:
-        f = open(log_name, 'w+')
-        data = []
-    data.append({'title': title, 'message': msg, 'result': res})
+    f = open(log_name, 'w+')
     f.write(json.dumps(data, indent=3))
     f.close()
```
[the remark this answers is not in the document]
grey-kristy: ok

cm/job.py
grey-kristy: It is not used now, but can be used in future. Just in case

cm/job.py
grey-kristy: yes, but I think it is more readable

cm/job.py
a-alferov: Ok.

cm/job.py
a-alferov: Ок

PR #72, opened 2019-12-26
Tests: Add functional tests for service and host removing

smuarenadata:
Host: If host has component, than API error.
      If host has NO component, then host deleted.
Service:
    If service has component mapped to host, than API error
    If service has active exports (bind), than API error
    If service has NO component mapped to host, than host should be deleted.


tests/functional/test_delete_host.py
```diff
@@ -0,0 +1,18 @@
...
+from adcm_client.objects import ADCMClient
+from adcm_pytest_plugin.utils import get_data_dir
+
+
+def test_delete_host(sdk_client_fs: ADCMClient):
+    """If host has NO component, than we can simple remove it from cluster.
+
+    :return:
+    """
+    hostprovider_bundle = sdk_client_fs.upload_from_fs(get_data_dir(__file__) + '/hostprovider')
```
acmnu: ```get_data_dir(__file__) + '/hostprovider'```

You can use ```get_data_dir(__file__, 'hostprovider')```

tests/functional/test_delete_host_data/hostprovider/config.yaml
```diff
@@ -0,0 +1,19 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+- type: provider
+  name: sample_provider
+  version: &version 0.4
+
+- type: host
+  name: vHost
+  version: 00.09
```
acmnu: Version should be string. We intend to force the restriction in a future.

tests/functional/test_delete_service_with_active_export.py
```diff
@@ -0,0 +1,51 @@
+# pylint: disable=W0611, W0621
```
acmnu: Missed license in file

PR #75, opened 2019-12-27
ADCM 1039 Added cancel tasks

a-alferov:


task_runner.py
```diff
@@ -11,19 +11,59 @@
...
+    jobs = JobLog.objects.filter(task_id=TASK_ID)
+
+    i = 0
+    while i < 10:
+        if jobs.filter(status=config.Job.RUNNING):
+            terminate_job(task, jobs)
+            break
+        i += 1
+        time.sleep(0.5)
+
```
grey-kristy: we need to call cm.job.finish_task() here in case we didn't find any running job

a-alferov: Ok, thanks.

PR #77, opened 2019-12-27
ADCM-1061 add rsync and adcm-client packages

amleshkov:


requirements.txt
```diff
@@ -20,3 +20,4 @@ apache-libcloud
 jmespath
 lxml
 pycrypto
+adcm-client
```
Maunty: all python packages that are needed for bundle work must be packed inside bundle tarball.
There is no documentation :) But take a look to https://jira.arenadata.io/browse/AP-30. 

```yaml
---

version: "1.0"
editions:
  - name: None
    exclude:
      - tests/
      - tests\.sh      
      - build/
      - _config.yaml
      - pylintrc
      - (.*/)?(\..+)
      # - .*pycache.*
      - .*README\.md
      - spec.yaml
      - requirements.yml

    preprocessors:
      - type: script
        script: build/config_builder.py
        args:
          - _config.yaml
      - type: python_mod_req
        requirements: requirements.yml
```

where second list element in preprocess will pack to bundle your packages from requirements.yml
To test local use adcm_sdk_pack shell util from adcm_client pack. 

requirements.yml format

```yaml
system_pkg:
  - musl-dev
  - linux-headers
  - gcc
python_mod:
  - openstacksdk

```
where system_pkg are system utils that are needed to be installed in adcm container before python package installation. This key is not required.

amleshkov:
Thanks!

PR #98, opened 2020-01-15
ADCM-1078 Adding a termination ban for some actions.

a-alferov:


api/serializers.py
```diff
@@ -783,7 +784,18 @@ class TaskListSerializer(serializers.Serializer):
     status = serializers.CharField(read_only=True)
     start_date = serializers.DateTimeField(read_only=True)
     finish_date = serializers.DateTimeField(read_only=True)
+    to_terminatable = serializers.SerializerMethodField()
```
grey-kristy: Why to_terminatable? Should be just terminatable, like upgradable

a-alferov: Ok.

a-alferov: Fixed

PR #112, opened 2020-01-20
ADCM-1087 add ui_options field to action description in config

grey-kristy:


acmnu:
bender build

Maunty:
Bender build

PR #113, opened 2020-01-20
Add libstdc++ package to base dockerfile

Maunty:
Yandex bundle needs yandex cloud client, which has grpc package in dependencies.
Grpc contains an SO which is dynamycaly linked with libstdc++.so


acmnu:
Close in favor to https://github.com/arenadata/adcm/pull/117

PR #133, opened 2020-01-27
ADCM-1104 add component prototype_id to component api output

grey-kristy:


api/cluster_serial.py
```diff
@@ -423,13 +423,17 @@ def get_kwargs(self, obj):
...
+    component_prototype_id = serializers.SerializerMethodField()
     display_name = serializers.SerializerMethodField()
     description = serializers.SerializerMethodField()
     url = MyUrlField(read_only=True, view_name='service-component-details')
 
     def get_name(self, obj):
         return obj.component.name
 
+    def get_component_prototype_id(self, obj):
+        return obj.component.id
```
a-alferov: Maybe we need `return obj.component.prototype.id` ?

grey-kristy: For what ? component.prototype.id is same as service.prototype.id in /api/v1/cluster/1/service/1/

a-alferov: Ok

PR #146, opened 2020-01-30
Test adcm 1068

smuarenadata:


tests/functional/test_upgrade_hostprovider.py
```diff
@@ -19,6 +19,31 @@
 from tests.library.errorcodes import UPGRADE_ERROR
 
 
+def test_check_prototype(sdk_client_fs: ADCMClient):
+    """Check prototype for service and cluster after upgrade
```
acmnu: > Check prototype for service and cluster after upgrade

provider and host

tests/functional/test_upgrade_cluster.py
```diff
@@ -19,6 +19,54 @@
 from tests.library.errorcodes import UPGRADE_ERROR
 
 
+def test_upgrade_with_two_clusters(sdk_client_fs: ADCMClient):
```
acmnu: What about two Provides? And both of providers should have a couple Host each.

tests/functional/test_upgrade_cluster_imports.py
```diff
@@ -0,0 +1,180 @@
...
+    cluster_import.bind(cluster)
+    upgr = cluster_import.upgrade(name='upgrade to 1.6')
+    id_before = cluster_import.prototype_id
+    upgr.do()
+    cluster_import.reread()
+    assert cluster_import.prototype().version == '1.6'
+    assert cluster_import.prototype_id != id_before
+
+
+def test_incorrect_import_strcit_version(sdk_client_fs: ADCMClient):
```
acmnu: I think it is simple to parametrize that test with bundle dirs for all variants of tests with <, <=, >, >=. You just need a copy of the same bundle.

PR #171, opened 2020-02-07
ADCM-478 Add test for host add/del plugins

acmnu:


acmnu:
Case with adcm_add_host and adcm_state will be added in another pull

PR #183, opened 2020-02-12
ADCM-1126 Added fail and success message to adcm_check

a-alferov:


python/ansible/plugins/action/adcm_check.py
```diff
@@ -77,18 +77,35 @@ def run(self, tmp=None, task_vars=None):
 
         result = super(ActionModule, self).run(tmp, task_vars)
```
grey-kristy: this string should be removed, it is no longer nessesary

a-alferov: Ok

PR #212, opened 2020-03-02
ADCM-1153 Log refactoring

a-alferov:


python/api/serializers.py
```diff
@@ -778,6 +782,27 @@ def get_content(self, obj):
...
+
+    def get_body(self, obj):
+        body = obj.body
+
+        if body is None and obj.name == 'ansible':
+            path_file = os.path.join(
+                config.RUN_DIR, f'{obj.job.id}', f'{obj.name}-{obj.type}.{obj.format}')
+            with open(path_file, 'r') as f:
+                body = f.read()
+
```
grey-kristy: Here we can return check log interactively: if body is None get data from CheckLog table

a-alferov:
Bender build

python/cm/migrations/0048_auto_20200228_1032.py
```diff
@@ -0,0 +1,96 @@
...
+    else:
+        return None
+
+
+def add_logs(apps, schema_editor):
+    LogStorage = apps.get_model('cm', 'LogStorage')
+    JobLog = apps.get_model('cm', 'JobLog')
+    db_alias = schema_editor.connection.alias
+
+    jobs = JobLog.objects.using(db_alias).all()
```
grey-kristy: using(db_alias) is superfluously.
You can write just JobLog.objects.all()

See, for example https://github.com/arenadata/adcm/blob/develop/python/cm/migrations/0037_auto_20191120_1600.py

a-alferov: Ok

PR #223, opened 2020-03-07
Test adcm 1043

smuarenadata:
Refactoring for cluster functions and host functions tests.


tests/functional/test_cluster_functions.py
```diff
@@ -42,48 +45,53 @@ def fin():
 
 @pytest.fixture(scope="module")
 def client(adcm):
-    steps.upload_bundle(adcm.api.objects, BUNDLES + "cluster_bundle")
-    steps.upload_bundle(adcm.api.objects, BUNDLES + "hostprovider_bundle")
+    steps.upload_bundle(adcm.api.objects, get_data_dir(__file__, 'hostprovider_bundle'))
```
acmnu: Please do not use upload_bundle. Rewrite this function with sdk_client_ms and upload_from_fs function. You could find api.objects inside sdk_client_ms intance to return it.

tests/functional/test_cluster_functions.py
```diff
@@ -306,43 +328,46 @@ def test_add_unknown_host_to_cluster(self, client):
             err.HOST_NOT_FOUND.equal(e, 'host doesn\'t exist')
         steps.delete_all_data(client)
 
-    def test_add_host_in_two_diff_clusters(self, client):
-        prototype = utils.get_random_cluster_prototype(client)
+    def test_add_host_in_two_diff_clusters(self, sdk_client_fs: ADCMClient):
+        bundle = sdk_client_fs.upload_from_fs(get_data_dir(__file__, 'cluster_simple'))
+        bundle_hp = sdk_client_fs.upload_from_fs(get_data_dir(__file__, 'hostprovider_bundle'))
         with allure.step('Create cluster1'):
```
acmnu: There is no need for that, bundle.cluster_create already do this message.

Actually there are a lot of places in adcm_client where we have allure decorators. Check out adcm_client code to find them.

PR #263, opened 2020-03-25
ADCM-1194 Added log rotation

a-alferov:


conf/adcm/config.yaml
```diff
@@ -36,3 +36,19 @@
         - name: "whitelisted_domains"
           type: list
           required: false
+    - name: "job_log"
+      display_name: "Job Log"
+      type: "group"
+      subs:
+        - name: "log_rotation_on_fs"
```
acmnu: Please add display_names for thus parameters.

python/task_runner.py
```diff
@@ -31,12 +34,43 @@
...
+def log_rotation():
+    adcm_object = ADCM.objects.get(id=1)
+    config_log = ConfigLog.objects.filter(obj_ref=adcm_object.config).latest('date')
+    adcm_config = json.loads(config_log.config)
+    log_rotation_on_db = adcm_config['job_log']['log_rotation_in_db']
+    log_rotation_on_fs = adcm_config['job_log']['log_rotation_on_fs']
+    rotation_jobs_on_db = []
+    rotation_jobs_on_fs = []
+
+    for job in JobLog.objects.all():
```
acmnu: I believe you could do a filter over finish_date with order_by and create index on finish_date. That allow you to get old much faster that iterate over all list. 

Your code now tend to be slow, because mostly this code iterate over thousands row to remove only one.

python/task_runner.py
```diff
@@ -31,12 +34,43 @@
...
+
+        if log_rotation_on_db > 0:
+            if finish_date < timezone.now() - timedelta(days=log_rotation_on_db):
+                rotation_jobs_on_db.append(job.id)
+
+        if log_rotation_on_fs > 0:
+            if finish_date < timezone.now() - timedelta(days=log_rotation_on_fs):
+                rotation_jobs_on_fs.append(job.id)
+
+    LogStorage.objects.filter(job_id__in=rotation_jobs_on_db).delete()
```
acmnu: Is it fast enought?

python/task_runner.py
```diff
@@ -137,6 +171,8 @@ def run_task(task_id, args=None):
 
     log.info("finish task #%s, ret %s", task_id, res)
 
+    log_rotation()
```
acmnu: What about parallel execution of that code. Will it survive?

os/etc/service/background_tasks/run
```diff
@@ -0,0 +1,19 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source /etc/adcmenv
+
+waitforwsgi
+
+echo "Run process_tasks ..."
+python "${adcmroot}/python/manage.py process_tasks --log-std"
```
acmnu: >> --log-std

It should be log file, not stdout

a-alferov: Yes, see to doc string, --log-std Redirect stdout and stderr to the logging system

python/cm/job.py
```diff
@@ -881,4 +888,37 @@ def run_task(task, event, args=''):
...
+
     set_task_status(task, config.Job.RUNNING, event)
+
+
+@background(schedule=1)
+def log_rotation():
+    log.info('Run log rotation')
+    adcm_object = ADCM.objects.get(id=1)
+    config_logs = ConfigLog.objects.filter(obj_ref=adcm_object.config)
+    adcm_conf = {}
```
grey-kristy, line 900: cl = ConfigLog.objects.get(obj_ref=adcm_object.config, id= adcm_object.config.current)
adcm_conf = json.loads(cl.config)

a-alferov: Fixed. https://github.com/arenadata/adcm/pull/281

PR #273, opened 2020-03-30
Tests for configuration fields

smuarenadata:
Tests, config generator.


acmnu:
Ok, I see no reason not to merge this one. It is work well and I think they do its job, so I intend to finally merge it.

PR #279, opened 2020-04-02
ADCM-1216 import test settings from work settings

grey-kristy:


tests/base/test_api.py
```diff
@@ -147,22 +147,22 @@ def test_access(self):
         api = ['cluster', 'host', 'job', 'task', 'stack', 'user', 'profile']
         for path in api:
             r1 = requests.get(self.url + '/' + path + '/')
-            self.assertEqual(r1.status_code, 403)
+            self.assertEqual(r1.status_code, 401)
```
acmnu, line 150: Why this change is here? No information in commit message about this.

PR #280, opened 2020-04-03
ADCM-1200 Added group parameters to adcm_check action plugin

a-alferov:


python/ansible/plugins/action/adcm_check.py
```diff
@@ -89,22 +130,46 @@ def run(self, tmp=None, task_vars=None):
...
+            'msg': msg,
+            'fail_msg': fail_msg,
+            'success_msg': success_msg,
+            'group_title': group_title,
+            'group_success_msg': group_success_msg,
+            'group_fail_msg': group_fail_msg
+        }
+
+        log.debug('ansible adcm_check: %s',
+                  ', '.join([f'{k}: {v}' for k, v in log_data.items() if v]))
```
grey-kristy, line 156: Lets move all this group logic into cm.job.log_check(). Lets call create_group_log() and log_group_check() from log_check()

If it requires too many params for log_check(), you could group them to some dict/dicts

a-alferov: Ok

PR #288, opened 2020-04-06
ADCM-1196 Fixed error 500 and refactoring try/except

a-alferov:


python/api/api_views.py
```diff
@@ -73,6 +73,15 @@ def select_serializer(self, request):
...
 
 
+def getlist_from_querydict(query_params, field_name):
+    params = query_params.get(field_name, [])
+    if params:
+        fields = [param.strip() for param in params.split(',')]
+    else:
+        return params
+    return fields
+
```
grey-kristy, line 83: Readability is not very good. I spent 5 munites to read this code.
Python way would be this:

```
if not params:
   return []
return [param.strip() for param in params.split(',')]
```

a-alferov: Ok

PR #334, opened 2020-04-27
ADCM-1194 Fixed rotation log from file system

a-alferov:


python/cm/job.py
```diff
@@ -953,14 +953,14 @@ def log_rotation():
...
     if log_rotation_on_fs:
-        rotation_jobs_on_fs = JobLog.objects.filter(
-            finish_date__lt=timezone.now() - timedelta(days=log_rotation_on_fs)).values('id')
-
-        if rotation_jobs_on_fs:
-
-            for job in rotation_jobs_on_fs:
-                shutil.rmtree(os.path.join(config.RUN_DIR, str(job['id'])))
-            log.info('rotation log from fs')
+        for folder in os.listdir(config.RUN_DIR):
```
acmnu, line 956: Do you understand that it could be not only folders in RUN_DIR?

python/cm/job.py
```diff
@@ -953,14 +953,14 @@ def log_rotation():
...
+            if not folder.startswith('.'):
```
acmnu, line 957: What this code for? It is not obvious why we have hidden files in RUN_DIR. Please made a comment about.

a-alferov: In the RUN_DIR folder is .gitignore

python/cm/job.py
a-alferov: And what else could be?

acmnu: File, regular file. You change logic, and create a lot of dirs in run, but on old installation there are a lot of files there.

python/cm/job.py
acmnu: It should not be there. Supposed we have to fix this in dockerignore file. Please remove that code and add exception to dockerignore

a-alferov: Ok

python/cm/job.py
a-alferov: If I understand correctly, you want me to add the deletion of files from the old log storage location LOG_DIR?

python/cm/job.py
a-alferov: code `if not folder.startswith('.'):` is used for development

python/cm/job.py
a-alferov: Added deletion of both folders and files

PR #356, opened 2020-05-08
ADCM-1271 Fixed 500 error for custom log, using invalid json

a-alferov:


python/api/job_serial.py
```diff
@@ -90,8 +90,11 @@ def get_content(self, obj):
                 content = json.loads(content)
         elif obj.type == 'custom':
             if obj.format == 'json' and isinstance(content, str):
-                content = json.loads(content)
-                content = json.dumps(content, indent=4)
+                try:
+                    content = json.loads(content)
```
acmnu, line 94: Is any chance for JSONDecodeError to be raised on 94 line? If so, after pass we will  have non text object at return time. 

It is way more safe to  use another variable to save result of json.loads.

a-alferov: Ok

PR #428, opened 2020-06-12
Test adcm 1312 service remove plugin

smuarenadata:


smuarenadata:
> You forget to check that it is possible to remove service with components mapped to hosts.

smuarenadata:
> You forget to check that it is possible to remove service with components mapped to hosts.

Fixed

PR #432, opened 2020-06-17
Refactoring for flaky tests.

smuarenadata:


tests/ui_tests/test_activatable_groups_ui_options.py
```diff
@@ -41,10 +41,11 @@ def login_on_adcm(app):
     __file__,
     "group_advanced_false_invisible_false_field_advanced_false_invisible_false_activiatable_false")
 def test_group_advanced_false_invisible_false_field_advanced_false_invisible_false_active_false(
-        sdk_client_ms: ADCMClient, path, app, login_on_adcm, gather_logs, screenshot_on_failure):
+        path, app, login_on_adcm, gather_logs, screenshot_on_failure):
     """Check that group not active and field is invisible until group is not active.
     """
     _ = login_on_adcm, gather_logs, screenshot_on_failure
+    sdk_client_ms = ADCMClient(app.adcm.api)
```
acmnu, line 48: Why????

smuarenadata: I want to try this changes because we have the same name for fixture and we have ADCM object creation in fixtures for tests, maybe sometimes we have duplicated ADCM or something like that that's why sometimes test can be failed, also at current moment we have this error:
```

AttributeError: 'Function' object has no attribute 'rep_call'

  File "/usr/local/lib/python3.6/dist-packages/allure_commons/_allure.py", line 169, in __call__
    return self._fixture_function(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/_pytest/fixtures.py", line 795, in _teardown_yield_fixture
    next(it)
  File "/adcm/tests/conftest.py", line 67, in app_fs
    if request.node.rep_call.failed:

````
I expect that this changed should fix it.

acmnu:
There is no need to do this. All fixture has good dependencies. So I close that PR.

PR #460, opened 2020-07-07
ADCM-1354 add script to backup sqlite db on every Django migrations

grey-kristy:


python/backupdb.py
```diff
@@ -0,0 +1,63 @@
...
+    except ImproperlyConfigured:
+        # No databases are configured (or the dummy one)
+        return False
+    if executor.migration_plan(executor.loader.graph.leaf_nodes()):
+        return True
+    return False
+
+
+def backup_sqlite(dbfile):
+    dt = datetime.datetime.now().strftime("%Y%m%d_%H%M")
```
acmnu, line 41: That could be not enough to limit this by minutes. Please add seconds to name.

acmnu: That could be a trouble in case of container restart during startup.

PR #471, opened 2020-07-13
ADCM-1350 unlock cluster after service deleting in adcm_delete_service ansible plugin

grey-kristy:


python/cm/job.py
```diff
@@ -258,7 +258,17 @@ def lock_objects(obj, event):
...
 
-def unlock_objects(obj, event):
+def unlock_deleted_objects(job, event):
+    if not job:
+        log.warning('unlock_deleted_objects: no job')
+        return
+    selector = json.loads(job.selector)
+    if 'cluster' in selector:
+        cluster = Cluster.objects.get(id=selector['cluster'])
+        unlock_obj(cluster, event)
```
a-alferov, line 268: Maybe instead of `unlock_obj(cluster, event)` call `unlock_objects(cluster, event)` to unlock other services and hosts for the cluster?

grey-kristy: yep

PR #487, opened 2020-07-27
ADCM-1339 Variant config type builtin functions

grey-kristy:


python/cm/adcm_config.py
```diff
@@ -342,33 +342,94 @@ def group_is_activatable(spec):
...
+    return out
+
+
+def variant_host_in_cluster(obj, args):
+    out = []
+    cluster = get_cluster(obj)
+    if not cluster:
+        return []
+
+    if 'service' in args:
```
a-alferov, line 387: if args = None, then there will be TypeError: argument of type 'NoneType' is not iterable

grey-kristy: thanks!

acmnu (changes requested):

acmnu (approved):

PR #503, opened 2020-08-06
ADCM-1378 Added JSONField for models

a-alferov:


python/cm/models.py
```diff
@@ -32,6 +33,38 @@
 )
 
 
+class JSONFieldFormatError(Exception):
```
grey-kristy, line 36: We already has custom error class AdcmEx and don't need another one. Just add new error code JSON_DB_ERROR, it would be enough. So we do not have to catch new exception through all the code.

a-alferov: Ok

python/cm/errors.py
```diff
@@ -123,6 +123,7 @@
     'NOT_IMPLEMENTED': ("not implemented yet", rfs.HTTP_501_NOT_IMPLEMENTED, ERR),
     'NO_JOBS_RUNNING': ("no jobs running", rfs.HTTP_409_CONFLICT, ERR),
     'BAD_QUERY_PARAMS': ("bad query params", rfs.HTTP_400_BAD_REQUEST),
+    'JSON_DB_ERROR': ("Not correct field format", rfs.HTTP_400_BAD_REQUEST, ERR),
```
grey-kristy, line 126: HTTP_409_CONFLICT

a-alferov: Fixed

grey-kristy (approved):

PR #508, opened 2020-08-10
ADCM-1382 Use JSON field type for model Component

a-alferov:


python/api/cluster_serial.py
```diff
@@ -496,7 +490,7 @@ def get_requires(self, obj):
         comp_list = {}
 
         def process_requires(req_list):
-            for c in json.loads(req_list):
+            for c in req_list:
```
acmnu, line 493: According to model req_list could be None. I think we should change blank=True in model and set default value to [].

python/cm/models.py
```diff
@@ -201,9 +201,9 @@ class Component(models.Model):
     name = models.CharField(max_length=160)
     display_name = models.CharField(max_length=160, blank=True)
     description = models.TextField(blank=True)
-    params = models.TextField(blank=True)       # JSON
-    constraint = models.TextField(blank=True)   # JSON
-    requires = models.TextField(blank=True)     # JSON
+    params = JSONField(blank=True)
```
acmnu, line 204: I don't like this blank=True. This forces us to do a lot of None check in functions. I think, that we should change it to blank=False and set reasonable defaults ([] for example). See my comments above about requires fields.

python/cm/stack.py
```diff
@@ -254,12 +254,15 @@ def save_components(proto, conf):
         dict_to_obj(cc, 'description', component)
         dict_to_obj(cc, 'display_name', component)
         dict_to_obj(cc, 'monitoring', component)
-        dict_json_to_obj(cc, 'params', component)
         fix_display_name(cc, component)
         check_component_constraint_definition(proto, comp_name, cc)
         check_component_requires(proto, comp_name, cc)
-        dict_json_to_obj(cc, 'constraint', component)
-        dict_json_to_obj(cc, 'requires', component)
+        if in_dict(cc, 'params'):
```
grey-kristy, line 260: Just use

dict_to_obj(cc, 'params', component)

It's one string instead of two :)

a-alferov: Ok

acmnu (approved):

PR #522, opened 2020-08-21
ADCM-1380 check required import issue for services

grey-kristy:
add tests for required import issue


python/cm/issue.py
```diff
@@ -188,27 +197,35 @@ def check_required_services(cluster):
...
+    log.debug('do_check_import result: %s, code: %s', res, code)
+    return res
+
+
+def do_check_import(cluster):
     def check_import(pi):
         if not pi.required:
-            return True
-        import_exist = False
+            return (True, 'NOT_REQIURED')
```
a-alferov, line 208: Brackets are not needed here

grey-kristy: It's more readable this way

a-alferov (approved):

PR #530, opened 2020-08-26
ADCM-1396 let adcm_config ansible module to change config of one service

grey-kristy:
from context of another service


a-alferov (approved):

acmnu (changes requested):
Stop working on ticket for a while

PR #555, opened 2020-09-10
ADCM-1413 add binded component constraint

grey-kristy:


python/cm/issue.py
```diff
@@ -252,6 +253,32 @@ def check_component_req(service, component):
...
+        binded_hc = binded_host_components(service, comp_name)
+        if not binded_hc:
+            msg = f'binded service "{service}", component "{comp_name}" not in hc for {ref}'
+            err('COMPONENT_CONSTRAINT_ERROR', msg)
+        for shc in binded_hc:
+            if not component_on_host(component, shc[1]):
+                msg = 'No binded component "{}" on host "{}" for {}'
+                err('COMPONENT_CONSTRAINT_ERROR', msg.format(component.name, shc[1].fqdn, ref))
+
+    for shc in [i for i in shc_list if i[2].component.binded_to]:
```
a-alferov, line 278: I think this is an excessive use of list comprehension, it would be better to put this `[i for i in shc_list if i[2].component.binded_to]` in a separate variable

python/cm/issue.py
```diff
@@ -247,26 +247,59 @@ def check_hc(cluster):
...
         check_component_requires(shc_list)
+        check_binded_components(shc_list)
     except AdcmEx:
         return False
     return True
 
 
 def check_component_requires(shc_list):
+    def get_components_with_requires():
+        return [i for i in shc_list if i[2].component.requires]
```
a-alferov, line 258: I suggest changing the name "i" to "shc", because "i" is a structure, not a number

acmnu (changes requested):
Test required

a-alferov (approved):

acmnu (approved):

PR #563, opened 2020-09-15
ADCM-1438 Config DRF refactoring

a-alferov:


python/api/config/serializers.py
```diff
@@ -0,0 +1,111 @@
...
+class HistoryCurrentPreviousConfigSerializer(serializers.Serializer):
+    history = serializers.SerializerMethodField()
+    current = serializers.SerializerMethodField()
+    previous = serializers.SerializerMethodField()
+
+    def get_history(self, obj):
+        view_name = 'config-history'
+        return ConfigURL(read_only=True, view_name=view_name).get_url(
+            obj, view_name, self.context['request'], format=None)
+
```
grey-kristy, line 53: You don't need to do it so complicated and use SerializerMethodField. Just write:

history = ConfigURL(read_only=True, view_name=view_name)

python/api/config/views.py
```diff
@@ -0,0 +1,147 @@
...
+    serializer_class = serializers.HistoryCurrentPreviousConfigSerializer
+    object_type = None
+
+    def get_queryset(self):
+        return get_objects_for_config(self.object_type)
+
+    def get(self, request, *args, **kwargs):
+        object_type = kwargs['object_type']
+        self.object_type = object_type
+        object_id = kwargs[f'{object_type}_id']
```
grey-kristy, line 73: This 3 string repeated too much. I suggest to write a function

object_type, object_id = get_obj_type_and_id(kwargs)

grey-kristy (changes requested):

python/api/config/serializers.py
a-alferov: Fixed in [https://github.com/arenadata/adcm/pull/563/commits/2ff1cbcf6ee5573650cd2fdf0928c4ac22360223](url)

python/api/config/views.py
a-alferov: Fixed in https://github.com/arenadata/adcm/pull/563/commits/2ff1cbcf6ee5573650cd2fdf0928c4ac22360223

grey-kristy (approved):

nagarin (approved):

PR #571, opened 2020-09-21
ADCM-1439 Added endpoint for service

a-alferov:


python/api/service/urls.py
```diff
@@ -0,0 +1,74 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from django.urls import path, include
+from . import views
+
+
+urlpatterns = [
+    path(
```
grey-kristy, line 19: I recommend new way to define url, something like this:

urlpatterns = [
    path('user/', include([
        path('', views.UserList.as_view(), name='user-list'),
        path('<name:username>/', include([
            path('', views.UserDetail.as_view(), name='user-details'),
            path('role/', views.ChangeUserRole.as_view(), name='change-user-role'),
            path('group/', views.AddUser2Group.as_view(), name='add-user-group'),
            path('password/', views.UserPasswd.as_view(), name='user-passwd'),
        ]))
    ])),
]

I think it is more readable and less redundant

a-alferov: I think this is a bad idea as we are limited by line length and word wrap will ruin readability.

a-alferov: Fixed

grey-kristy (approved):

acmnu:
bender build

PR #587, opened 2020-10-06
ADCM-1394 fix(status) URL to request status

DavidovPavel:


web/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve",
+    "start": "ng serve --proxyConfig=proxy.conf.json",
```
acmnu, line 7: Is this a mistake?

acmnu (changes requested):

web/package.json
DavidovPavel: No

acmnu (approved):

PR #598, opened 2020-10-19
ADCM-1435 move adcm credentials to separate fixture

skhomuti:


tests/conftest.py
```diff
@@ -119,11 +119,24 @@ def app_fs(adcm_fs: ADCM, web_driver: ADCMTest, request):
...
+@pytest.fixture(scope='session')
+def adcm_credentials():
+    """
+    Provides ADCM username and password by default
+    Examples:
+        login, password = adcm_credentials
+        ...
+        login(*adcm_credentials)
+    """
+    yield 'admin', 'admin'
```
dgusakov, line 131: Since you are using *tuple syntax maybe it will be better (in terms for further usage) to return dict here?
```suggestion
    return {"username" = "admin", "password" = "admin"}
```
and use **dict syntax

Also yield makes no sense here

skhomuti: I thought about it. The problem is that different functions have different argument names.
```
def login(self, login, password):
...
def auth(self, username, password):
```
probably i can rename login to username, if necessary

dgusakov (approved):

vgorkavenko (approved):

PR #669, opened 2020-12-14
ADCM-1486 Added tests for upgrade object config

a-alferov:


python/cm/tests_upgrade.py
```diff
@@ -291,6 +297,794 @@ def test_from_active_group_to_not_active_group(self):
...
+        old_conf, old_attr = get_config(cluster)
+
+        self.assertEqual(old_conf, {'advance': {'port': 11}})
+        self.assertEqual(old_attr, {})
+
+        adcm_config.switch_config(cluster, p2, p1)
+
+        new_conf, new_attr = get_config(cluster)
+        print(f'conf: {new_conf}\nattr: {new_attr}')
+        self.assertTrue(False)
```
skhomuti, line 350: tests don't look complete

skhomuti (changes requested):

python/cm/tests_upgrade.py
a-alferov: Yes, that is right

grey-kristy (approved):

acmnu:
Let's close it in favor to future better tests

PR #749, opened 2021-02-15
Black formatting for tests

dgusakov:


acmnu (changes requested):
Let's do it in the same time for tests and for main code. It is better when we have one huge commit instead of two.

PR #750, opened 2021-02-16
ADCM-1599 FIX Internal server error for ADS bundle upload

a-alferov:


python/cm/bundle.py
```diff
@@ -254,7 +254,7 @@ def re_check_actions():
 def check_component_requires(comp):
     if not comp.requires:
         return
-    ref = 'in requires of component "{}" of {}'.format(comp.name, proto_ref(comp.prototype))
+    ref = f'in requires of {comp.type} "{comp.name}" {comp.version}'
```
grey-kristy, line 257: No,  I think this

`ref = 'in requires of component "{}" of {}'.format(comp.name, proto_ref(comp.parent))`

would be better

grey-kristy (changes requested):

python/cm/bundle.py
a-alferov: Fixed

grey-kristy (approved):

PR #759, opened 2021-02-19
Solver for host builtin function of variant config type

grey-kristy:


python/cm/variant.py
```diff
@@ -0,0 +1,319 @@
...
+    else:
+        return None
+    return cluster
+
+
+def variant_service_in_cluster(obj, args=None):
+    out = []
+    cluster = get_cluster(obj)
+    if not cluster:
+        return []
```
a-alferov, line 35:
```suggestion
        return out
```
You can return out instead of an empty list

python/cm/variant.py
```diff
@@ -0,0 +1,319 @@
...
+    for co in ClusterObject.objects.filter(cluster=cluster).order_by('prototype__name'):
+        out.append(co.prototype.name)
+    return out
+
+
+def variant_service_to_add(obj, args=None):
+    out = []
+    cluster = get_cluster(obj)
+    if not cluster:
+        return []
```
a-alferov, line 46:
```suggestion
        return out
```
You can return out instead of an empty list

python/cm/variant.py
```diff
@@ -0,0 +1,319 @@
...
+        cluster = obj
+    else:
+        return None
+    return cluster
+
+
+def variant_service_in_cluster(obj, args=None):
+    out = []
+    cluster = get_cluster(obj)
+    if not cluster:
```
a-alferov, line 34:
```suggestion
    if cluster is None:
```
I think it would be better to explicitly check for None, since the `get_cluster()` function returns an object or None

python/cm/variant.py
```diff
@@ -0,0 +1,319 @@
...
+
+    for co in ClusterObject.objects.filter(cluster=cluster).order_by('prototype__name'):
+        out.append(co.prototype.name)
+    return out
+
+
+def variant_service_to_add(obj, args=None):
+    out = []
+    cluster = get_cluster(obj)
+    if not cluster:
```
a-alferov, line 45:
```suggestion
    if cluster is None:
```
See comment above

python/cm/variant.py
```diff
@@ -0,0 +1,319 @@
...
+    if 'predicate' not in args:
+        err('CONFIG_VARIANT_ERROR', 'no "predicate" key in variant host function arguments')
+    res = var_host_solver(cluster, VARIANT_HOST_FUNC, args)
+    return res
+
+
+def variant_host_in_cluster(obj, args=None):
+    out = []
+    cluster = get_cluster(obj)
+    if not cluster:
```
a-alferov, line 230:
```suggestion
    if cluster is None:
```

python/cm/variant.py
```diff
@@ -0,0 +1,319 @@
...
+        return []
```
a-alferov, line 231:
```suggestion
        return out
```

python/cm/variant.py
```diff
@@ -0,0 +1,319 @@
...
+
+    if args and 'service' in args:
+        try:
+            service = ClusterObject.objects.get(cluster=cluster, prototype__name=args['service'])
+        except ClusterObject.DoesNotExist:
+            return []
```
a-alferov, line 237:
```suggestion
            return out
```

python/cm/variant.py
```diff
@@ -0,0 +1,319 @@
...
+        if 'component' in args:
+            try:
+                comp = ServiceComponent.objects.get(
+                    cluster=cluster, service=service, prototype__name=args['component']
+                )
+            except ServiceComponent.DoesNotExist:
+                return []
```
a-alferov, line 244:
```suggestion
                return out
```

a-alferov (approved):

PR #768, opened 2021-03-01
CI-749 Implement make pytest_release

julya935:


pytest.sh
```diff
@@ -20,7 +20,7 @@ find . -name "__pycache__" -type d -delete
 { # try
     pytest tests/ui_tests tests/functional -s -v -n auto --maxfail 30 \
     --showlocals --alluredir ./allure-results/ --durations=20 \
-    --reruns 2 --remote-executor-host "$SELENOID_HOST" --timeout=360 &&
+    --reruns 2 --remote-executor-host "$SELENOID_HOST" --timeout=360 --$1 &&
```
Maunty, line 23: move `--` inside of `$1`
also  `$@` will be better then `$1`. Mb in future there will be some more options along with `--firefox`

Maunty (changes requested):

Maunty (approved):

PR #777, opened 2021-03-05
ADCM-1633 Add tests for dependent components

madlabman:


tests/functional/test_related_sc_data/bundle_configs/A_X_to_A_Y/cases/case_1.yaml
```diff
@@ -0,0 +1,6 @@
+description: Attempt to save HC map with only A.X on 1 host
+bound_to_raise_exception:
+requires_raise_exception:
```
dgusakov, line 3: I don't think that this is the correct way of using YAML keys. Either pass them as list values or remove them at all and use existing logics from bound components tests.

tests/functional/test_related_sc_data/bundle_configs/A_X_to_A_Y/cases/case_4.yaml
```diff
@@ -1,5 +1,5 @@
-description: Attempt to save HC map with A.X and A.Y on first host and A.Y on second host- should fail
-positive: False
+description: Attempt to save HC map with A.X and A.Y on first host and A.Y on second host
```
dgusakov, line 1: Information about test expectations is useful. Especially in the test description

dgusakov (changes requested):
Separate tests for related and bound components.

Why some configs were removed? Bounded and required are two completely different logics. So mixing it in one test with combined data is not a good idea.

tests/functional/test_related_sc.py
```diff
@@ -49,29 +47,32 @@ def get_cases_paths() -> CasesPathsForParametrize:
...
+    https://arenadata.atlassian.net/browse/ADCM-1633
     """
     with allure.step('Upload custom provider bundle and create it in ADCM'):
         provider_bundle = sdk_client_fs.upload_from_fs(get_data_dir(__file__) + "/provider")
         provider = provider_bundle.provider_prototype().provider_create(random_string())
     with allure.step('Upload custom cluster bundle and create it in ADCM'):
-        cluster_bundle = sdk_client_fs.upload_from_fs(case_path.split(CASES_PATH)[0])
+        bundle_path = os.path.join(case_path.split(CASES_PATH)[0], relation_type)
+        if not os.path.exists(os.path.join(bundle_path, "config.yaml")):
+            pytest.skip(f"bundle for relation type '{relation_type}' does not exist")
```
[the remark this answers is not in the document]
madlabman: Test cases has been split

tests/functional/test_related_sc_data/bundle_configs/A_X_to_A_Y/cases/case_4.yaml
madlabman: Fixed in split test cases

dgusakov (approved):

PR #788, opened 2021-03-16
ADCM-1530/ADCM-1642 Fixed the occurrence of service issue when updating the cluster

a-alferov:


grey-kristy (changes requested):
After upgrade we need to recalculate issue on all objects - cluster, service, component, provider and host, from down to top

python/cm/upgrade.py
```diff
@@ -256,6 +256,23 @@ def rpm_cmp_reverse(obj1, obj2):
         return res
 
 
+def re_count_issue(obj, upgrade):
+    """Re-count of issues from the bottom up"""
+    if obj.prototype.type == 'cluster':
+        for proto in Prototype.objects.filter(bundle=upgrade.bundle, type='service'):
+            for co in ClusterObject.objects.filter(cluster=obj, prototype=proto):
+                for comp in Prototype.objects.filter(parent=proto, type='component'):
```
acmnu, line 264: There is no needs to iterate over component's prototype. Filter from ServiceComponent.objects.filter(                            cluster=obj, service=co) is enough to got a component. 

In current realisation for every comp you will have only one sc and no way to got two or more sc in line 265. So, just remove loop from line 264

python/cm/upgrade.py
```diff
@@ -256,6 +256,23 @@ def rpm_cmp_reverse(obj1, obj2):
         return res
 
 
+def re_count_issue(obj, upgrade):
+    """Re-count of issues from the bottom up"""
```
acmnu, line 260: Or? It will recount from bottom for cluster and provider only. May be it is better to mark this in docstring? Besides it would be nice to have recount for service too, just in case.

python/cm/upgrade.py
```diff
@@ -256,6 +256,23 @@ def rpm_cmp_reverse(obj1, obj2):
         return res
 
 
+def re_count_issue(obj, upgrade):
```
acmnu, line 259: Recount is a single word.

acmnu (changes requested):

python/cm/upgrade.py
a-alferov: Fixed

python/cm/upgrade.py
a-alferov: Fixed

python/cm/upgrade.py
a-alferov: Fixed

python/cm/upgrade.py
```diff
@@ -256,6 +256,24 @@ def rpm_cmp_reverse(obj1, obj2):
         return res
 
 
+def recount_issues(obj, upgrade):
+    """
+    Recount of issues from the bottom up for all related objects component-service-cluster
+    or host-provider, or for cluster or provider objects
```
acmnu, line 262: >  or for cluster or provider objects

Incorrect. According to code you do recursion for cluster and provider and do no recursion for component, service and host.

acmnu (changes requested):

python/cm/upgrade.py
acmnu: Not fixed, see new comment

python/cm/upgrade.py
```diff
@@ -256,6 +256,24 @@ def rpm_cmp_reverse(obj1, obj2):
...
+    """
+    if obj.prototype.type == 'cluster':
+        for proto in Prototype.objects.filter(bundle=upgrade.bundle, type='service'):
```
grey-kristy, line 265: This loop is unnecessary. You can get services straight from ClusterObject

python/cm/upgrade.py
```diff
@@ -256,6 +256,24 @@ def rpm_cmp_reverse(obj1, obj2):
...
+            for co in ClusterObject.objects.filter(cluster=obj, prototype=proto):
+                for sc in ServiceComponent.objects.filter(cluster=obj, service=co):
+                    cm.issue.save_issue(sc)
+                cm.issue.save_issue(co)
+    elif obj.prototype.type == 'provider':
+        for proto in Prototype.objects.filter(bundle=upgrade.bundle, type='host'):
```
grey-kristy, line 271: This too

grey-kristy (changes requested):

python/cm/upgrade.py
a-alferov: Fixed

python/cm/upgrade.py
a-alferov: Fixed

python/cm/upgrade.py
a-alferov: Fixed

grey-kristy (approved):

acmnu (approved):

PR #789, opened 2021-03-16
Bump and assemble refactoring

acmnu:
This commit bump software to latest version. Main change is
Python 3.9. Minor part of this a lot of different small packages.

This commit adds rust and cargo as build time dependencies bescause
of new parts in crypto engine (PEP517).

This commit also change logic of assembling. Everything related to
build is in assemble dir as it was done for adss_core previously.


requirements.txt
```diff
@@ -1,24 +1 @@
...
-toml
-uwsgi
-version_utils
-yspec
-# Custom bundle libs
-apache-libcloud
-jmespath
-lxml
-pycrypto
+-r assemble/base/requirements-base.txt
```
grey-kristy, line 1: May be yspec should be here?

acmnu: I think it should be in separate PR

acmnu:
bender build

BenderArenadata:

grey-kristy (approved):

PR #791, opened 2021-03-18
ADCM-1617 Remove usages of api.objects in ADCM tests #2

madlabman:


tests/functional/test_cluster_functions.py
```diff
@@ -9,443 +9,249 @@
...
-            assert e.value.error['prototype_id'][0] == 'A valid integer is required.'
-        steps.delete_all_data(client)
-
-    def test_shouldnt_add_two_identical_service_in_cluster(self, sdk_client_fs: ADCMClient):
-        bundle = sdk_client_fs.upload_from_fs(get_data_dir(__file__, 'cluster_simple'))
-        cluster = bundle.cluster_create(utils.random_string())
+                _create_service(cluster, utils.random_string())
+        with allure.step("Check error about valid integer"):
+            assert e.value.error.title == "400 Bad Request"
+            assert e.value.error["prototype_id"][0] == "A valid integer is required."
```
dgusakov, line 231: This test should be moved to Unit tests too

madlabman: There is no cluster service tests in unit tests. Are you sure we have to move this test?

dgusakov: Since you have already started this activity, yes I'm sure.
@grey-kristy Seems like we need your help here with unit tests

madlabman: [done](https://github.com/arenadata/adcm/pull/791/commits/277da469151975789e8f92b7665507ddd0f1b3a2)

dgusakov (approved):

tests/functional/test_cluster_functions.py
```diff
@@ -9,443 +9,222 @@
...
-    def test_shouldnt_create_cluster_when_proto_is_string(self, client):
-        with pytest.raises(coreapi.exceptions.ErrorMessage) as e:
-            client.cluster.create(prototype_id=utils.random_string(), name=utils.random_string())
-        with allure.step('Check error about valid integer'):
-            assert e.value.error.title == '400 Bad Request'
-            assert e.value.error['prototype_id'][0] == 'A valid integer is required.'
-        steps.delete_all_data(client)
+def _create_service(cluster_: Cluster, prototype_id):
+    """Workaround to create service without retrieving prototype"""
+    cluster_._subcall(
```
a-alferov, line 67: Why do you use private methods, the cluster object has an service_add() method. Isn't it enough for you?

tests/functional/test_cluster_functions.py
```diff
@@ -9,443 +9,222 @@
...
-            assert e.value.error.title == '400 Bad Request'
-            assert e.value.error['description'] == ["This field may not be blank."]
-
-    def test_get_cluster_list(self, sdk_client_fs: ADCMClient):
-        bundle = sdk_client_fs.upload_from_fs(get_data_dir(__file__, 'cluster_simple'))
-        expectedlist = []
-        actuallist = []
+def _patch_cluster(cluster_: Cluster, **kwargs):
+    """Call partial_update endpoint (does not implemented in adcm_client)"""
+    cluster_._subcall("partial_update", **kwargs)
```
a-alferov, line 74: Why are you using private methods? If you need methods that are not in the client, then you need to ask us to add these methods, and not use private ones.

tests/functional/test_cluster_functions.py
madlabman: Some test cases require to set config directly because `config_set` at least expect iterable.

tests/functional/test_cluster_functions.py
madlabman: I'll create feature request and replace by added method after release.
All of the extracted methods with private methods calls are possible candidates to implement in adcm_client, expect some cases like in the comment above.

a-alferov (approved):

PR #799, opened 2021-03-23
Remove steps.py

madlabman:
Remove _client from ADCMTest class


tests/functional/test_adcm_check_plugin.py
```diff
@@ -26,6 +26,15 @@
                     "Task fail", False, False)]
 
 
+@allure.step("Check if action {action_name} state is {state_expected}")
+def check_action_state(action_name: str, state_current: str,
+                       state_expected: str) -> None:
+    assert state_current == state_expected, (
+        f"Current action {action_name} status {state_current}. "
+        f"Expected: {state_expected}"
+    )
```
dgusakov, line 35: We have https://github.com/arenadata/adcm_pytest_plugin/blob/release/3.x/src/adcm_pytest_plugin/steps/asserts.py

dgusakov (changes requested):

tests/functional/test_adcm_check_plugin.py
madlabman: Replaced with `assert_action_result`

dgusakov (approved):

PR #810, opened 2021-03-26
ADCM-1664 add component to adcm_state ansible plugin

grey-kristy:


python/cm/api.py
```diff
@@ -868,6 +867,35 @@ def set_host_state(host_id, state):
...
+        comp = ServiceComponent.objects.get(id=component_id)
+    except ServiceComponent.DoesNotExist:
+        msg = 'Component # {} does not exist'
+        err('COMPONENT_NOT_FOUND', msg.format(component_id))
+    return push_obj(comp, state)
+
+
+def set_component_state_by_name(cluster_id, service_id, component_name, service_name, state):
+    try:
+        if service_id:
```
a-alferov, line 881:
```suggestion
        if service_id is not None:
```
I think it's better to explicitly check on None, so as not to mask possible other problems

a-alferov (approved):

PR #815, opened 2021-03-31
Try new uniform error processing

grey-kristy:


python/cm/models.py
```diff
@@ -36,6 +37,25 @@
...
+        except ObjectDoesNotExist:
+            if not hasattr(self.model, '__error_code__'):
+                raise AdcmApiEx('NO_MODEL_ERROR_CODE', f'model: {self.model.__name__}') from None
+            msg = '{} {} does not exist'.format(self.model.__name__, kwargs)
+            raise AdcmApiEx(self.model.__error_code__, msg) from None
+
+
+class ADCMModel(models.Model):
+    objects = models.Manager()
+    obj = ADCMManager()
```
nagarin, line 53: Why not to just override default `objects` with custom manager, or there are a lot of code waiting for  standard `ObjectDoesNotExist`?

grey-kristy: Because I can't. If just replace standart manager with custom one, DRF and django won't start with a lot of errors

nagarin: I see, probably DRF depends on it a lot (or where should it get 404 errors).

PR #818, opened 2021-04-02
ADCM-1669 issue.py refactoring

nagarin:


python/cm/issue.py
```diff
@@ -10,18 +10,65 @@
...
+    Prototype,
+    PrototypeImport,
+    ServiceComponent,
+)
+
+
+def update_hierarchy_issues(obj):
+    save_issue(obj)
+
+    if not obj or obj.prototype.type == 'adcm':
```
a-alferov, line 34: This check is repeated in the `save_issue()` function (line 66-70). I think you can remove them from the `save_issue()` function, and move the check call to line 32.

a-alferov (changes requested):

python/cm/issue.py
```diff
@@ -114,54 +153,41 @@ def get_issue(obj):   # pylint: disable=too-many-branches
...
-    if not check_config(cluster):
-        issue['config'] = False
-    if not check_required_services(cluster):
-        issue['required_service'] = False
-    if not check_required_import(cluster):
-        issue['required_import'] = False
-    if not check_hc(cluster):
-        issue['host_component'] = False
-    return issue
+    return {
```
grey-kristy, line 165: It is the change of semantic. Now frontend assumed to get empty issue, if issue is absent. I think we should keep this behavior.

grey-kristy (changes requested):

python/cm/issue.py
nagarin: Nope, same logic remains, just moved in single place to line 119 of this file

python/cm/issue.py
```diff
@@ -10,18 +10,89 @@
...
+        result.append(obj)
+        result.extend(ServiceComponent.objects.filter(service=obj))
+
+    elif obj.prototype.type == 'component':
+        result.append(obj.service.cluster)
+        result.append(obj.service)
+        result.append(obj)
+
+    elif obj.prototype.type == 'host':
+        result.append(obj.cluster)
```
grey-kristy, line 70: Don't we need to add
 
`result.append(obj)`

here?

nagarin: No,  it's in other function (get_affected_host_hierarchy), same for cluster there

a-alferov (dismissed):

grey-kristy (approved):

PR #819, opened 2021-04-05
ADCM-1671 add new locked objects tests

skhomuti:


tests/functional/test_locked_objects.py
```diff
@@ -9,190 +9,307 @@
...
+    def test_lock_unlock(self, cluster, host):
+        """
+        Test that cluster locked when action running and unlocked when action ends
+        """
+        cluster.host_add(host)
+        assert_state(cluster, "created")
+        task = _lock_obj(cluster)
+        _check_that_object_is_locked(cluster)
+        task.wait()
+        assert_state(cluster, "created")
```
dgusakov, line 70: Here and below in we assert state and actions list to ensure that object was locked.
But when we check obj unlock we assert only state.
Does it enough? Or should we add actions list assertion for unlock obj check too?

skhomuti: I thought about it too

dgusakov: Let's add it

vgorkavenko (approved):

dgusakov (approved):

madlabman (approved):

PR #823, opened 2021-04-13
ADCM-1682 remove duplication of service code in api

grey-kristy:


a-alferov (commented):
You need to check filtering by the display_name field for services in https://github.com/arenadata/adcm/blob/adcm-1682/python/api/api_views.py#L188 Since the ClusterServiceList class has been removed.

a-alferov (approved):

PR #827, opened 2021-04-16
ADCM-1691 remove duplication of host code in api

grey-kristy:


python/api/cluster_serial.py
```diff
@@ -73,7 +74,7 @@ class ClusterDetailSerializer(ClusterSerializer):
     license = serializers.CharField(read_only=True)
     action = CommonAPIURL(view_name='object-action')
     service = ServiceURL(view_name='service')
-    host = hlink('cluster-host', 'id', 'cluster_id')
+    host = ServiceURL(view_name='host')
```
a-alferov, line 77: Why `ServiceURL,` maybe create `HostURL` or generic class like `HyperlinkedURL`

a-alferov (approved):

PR #861, opened 2021-05-27
try to use GenericForeignKey in TaskLog - first draft

grey-kristy:


grey-kristy:
convert to ADCM-1734 https://github.com/arenadata/adcm/pull/867

PR #864, opened 2021-05-28
Added abstract class for ADCM objects

a-alferov:


nagarin (changes requested):
Could we make models change in separate PR like this? I need the same base abstract model in my work already

nagarin (approved):

PR #870, opened 2021-06-03
ADCM-1739 Added `ConfigGroup` and `HostGroup` models, endpoints and CRUD for models

a-alferov:


python/cm/models.py
```diff
@@ -219,6 +250,12 @@ def serialized_issue(self):
 class Cluster(ADCMEntity):
     name = models.CharField(max_length=80, unique=True)
     description = models.TextField(blank=True)
+    config_groups = GenericRelation(
+        'ConfigGroup',
+        object_id_field='object_id',
+        content_type_field='object_type',
+        on_delete=models.CASCADE,
```
nagarin, line 257: Seems like it is not used - referenced objects are deleted with Cluster even when is set to SET_NULL, anyway it's meaningless since it should be set from ForeignKey side of relation

nagarin (changes requested):

python/cm/models.py
a-alferov: This is how GenericForeignKey works. We need to use CASCADE on GenericRelation so that groups are deleted along with the deleted object.

nagarin: OK, let it stay in sake of explicity

nagarin (approved):

grey-kristy (approved):

grey-kristy (approved):

PR #873, opened 2021-06-08
ADCM-1749 Implement regression tests for infinite host lock

vgorkavenko:


dgusakov (dismissed):

vgorkavenko:
Need to implement negative tests too

PR #882, opened 2021-06-15
ADCM-1762 Develop tests for Expand on locked host

vgorkavenko:


dgusakov:
@vgorkavenko Is there a way to make the error message `Failed: DID NOT RAISE <class 'coreapi.exceptions.ErrorMessage'>` more readable and verbose?

dgusakov (approved):

PR #899, opened 2021-06-28
Fix alpine version in buildss

pisohaka:
Fix alpine version in buildss target in Makefile: set image version to
golang:1.15-alpine3.13 because by defult image with alpine3.14 is used
and there is a bug with make in alpine3.14.


Maunty (approved):

pisohaka:
bender build

BenderArenadata:
Hey, @pisohaka! Build has been launched!

PR #935, opened 2021-07-14
 ADCM-1815 Fixed bug with page reload after opening from the toolbar

sanshan:


dgusakov:
@sanshan Issue ADCM-1815 from the PR name is not related to the code changes here. Pls consider changing issue ref to the  correct one

PR #940, opened 2021-07-19
ADCM-1911 Tests on adcm_config ability to change bundle parts configuration

Sealwing:
Vertical configuration changes


Sealwing:
random.choice is used to pick *one of* clusters/providers/services/hosts/components to change it alone, without affecting neighbors

dgusakov (approved):

PR #941, opened 2021-07-20
ADCM-1855 API based login in general UI tests

Sealwing:


Sealwing:
bender test

tests/conftest.py
```diff
@@ -211,6 +213,18 @@ def bundle_archive(request, tmp_path):
     return _pack_bundle(request.param, tmp_path)
 
 
+@pytest.fixture(scope="function")
```
dgusakov, line 216: Add allure.title decorator with human-readable fixture name

tests/ui_tests/test_numbers.py
```diff
@@ -20,7 +20,7 @@
 
 
 @parametrize_by_data_subdirs(__file__, 'bundles')
-def test_number_validation(sdk_client_fs: ADCMClient, path, app_fs, login_to_adcm):
+def test_number_validation(sdk_client_fs: ADCMClient, path, app_fs, api_auth_to_adcm):
```
dgusakov, line 23: Replace other occurrences of `login_to_adcm` fixture.
Also rename fixture `login_to_adcm` to `login_to_adcm_over_ui` for the further strict separation of `login_to_adcm_over_ui` and new fixture

tests/conftest.py
```diff
@@ -211,6 +213,18 @@ def bundle_archive(request, tmp_path):
...
+def api_auth_to_adcm(app_fs, adcm_credentials):
```
dgusakov, line 217: rename to `login_to_adcm_over_api`

dgusakov (changes requested):

tests/conftest.py
Sealwing: done

tests/ui_tests/test_numbers.py
Sealwing: done

dgusakov (approved):

tests/ui_tests/test_numbers.py
```diff
@@ -56,8 +56,8 @@ def test_number_in_range_values(
         assert config.save_button_status()
 
 
-def test_float_in_integer_field(sdk_client_fs: ADCMClient, app_fs, login_to_adcm):
-    _ = login_to_adcm
+def test_float_in_integer_field(sdk_client_fs: ADCMClient, app_fs, login_to_adcm_over_api):
+    _ = login_to_adcm_over_api
```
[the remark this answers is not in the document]
Sealwing: Who knows... Looks like pylint avoider, should be removed

dgusakov (dismissed):

madlabman (approved):

Sealwing:
Investigate fail with tests where Configuration class is in play
Before Merge

grey-kristy (approved):

dgusakov (approved):

Sealwing:
bender test

Sealwing:
bender build

BenderArenadata:
Hey, @Sealwing! Build has been launched!

tests/ui_tests/app/app.py
```diff
@@ -41,6 +41,8 @@ def __init__(self, browser="Chrome"):
         self.opts.add_argument("--start-maximized")
         self.opts.add_argument("--enable-logging")
         self.opts.add_argument("--enable-automation")
+        if browser == "Chrome":
+            self.opts.add_argument("--window-size=1366,768")
```
madlabman, line 45: So why we have [`set_window_size`](https://github.com/arenadata/adcm/blob/423004f3ae58a8e6dc947b87e720b566989a20d7/tests/ui_tests/app/app.py#L72) below?

Sealwing: @annademina
Do we need it now for some reason?
I didn't test it without `set_window_size` in non-headless mode, so not sure how does it affect debug

annademina: as I remember we decided to set this window size to check if there are any problems with the website layout? Not only because of popups and etc.

Sealwing: The thing is that we have 2 lines that setting window size:
- this one sets it via options (and works in headless mode)
- the one that @madlabman mentioned comes after this one and sets resolution via .set_window_size method that works in non-headless mode, but seems to have no effect on headless one, at least for Chrome

Maybe it works with Firefox (but as I know it's about 1366x768 by default in headless)

Sealwing: We'll test it against Firefox and remove it if it's required no more.
Just not as part of current PR.

madlabman: I suggest to keep the wider one (by driver options).

Sealwing: done.
added same for Firefox

madlabman (approved):

dgusakov (approved):

annademina (approved):

PR #943, opened 2021-07-21
ADCM-1880, ADCM-1882, ADCM-1886 provide specs + fixes

AKhodus:


acmnu:
Cancel this activity because everything goes to wiki now.

PR #947, opened 2021-07-23
ADCM-1934 Fix bug with unlimited username

SofiaTal:
Fix bug with too long user name on user managment page in adcm. Add Error message discribes users their mistake in case of extra long username.


python/api/user/serializers.py
```diff
@@ -65,14 +66,22 @@ class GroupDetailSerializer(GroupSerializer):
...
     change_password = hlink('user-passwd', 'username', 'username')
     change_role = hlink('change-user-role', 'username', 'username')
     is_superuser = serializers.BooleanField(required=False)
 
+    def validate_username(self, name):
+        if len(name) > MAX_NAME_LENGTH:
+            raise AdcmEx(
+                'LONG_NAME',
+                f'Username is too long, up to {MAX_NAME_LENGTH} symbols required',
+            )
```
dgusakov, line 82: Need to make this error readable at UI
Screenshot 2021-07-27 at 11 48 34

dgusakov (changes requested):

python/api/user/serializers.py
SofiaTal: Fixed in last commit

dgusakov (approved):
Tested. The bug is fixed.

PR #949, opened 2021-07-23
ADCM-1844_4 add network waiting

annademina:
What has been changed and why:
1) now a new tab is opened exactly for the test (there were versions with no new tab opening and opening a new tab on the stage of the previous test running).
2) added waiting for config. without this waiting, there was an unexpected redirection to the greeting page when this config is finally loaded.


tests/ui_tests/app/helpers/network.py
```diff
@@ -0,0 +1,77 @@
...
+        ))
+    return result
+
+
+def get_network_log(browser: WebDriver, is_web_doc=True) -> list:
+    network_data = fetch_requests(browser, is_web_doc)
+    return transform_network_log(network_data)
+
+
+@allure.step('Wait for requests to stop')
```
dgusakov, line 68:
```suggestion
@allure.step('Wait for all network requests to stop')
```

tests/ui_tests/app/helpers/network.py
```diff
@@ -0,0 +1,77 @@
...
+@contextmanager
+def wait_all_requests_stop(browser: WebDriver):
+    yield
+
+    def is_stopped():
+        requests_before = get_network_log(browser)
+        time.sleep(1)
+        assert requests_before == get_network_log(browser)
```
dgusakov, line 76: What if the request takes more than 1 second?

tests/ui_tests/app/helpers/network.py
```diff
@@ -0,0 +1,77 @@
...
+            domain=domain,
+            method=method,
+            status=status,
+            status_text=status_text,
+            duration=duration,
+            type=type,
+            url=url,
+            referer=referer
+        ))
+    return result
```
dgusakov, line 60: This code is hardly needed here you add it for some upcoming features? It looks like now it will be faster and easier just to assert a number of lines in the log and not parse it each time

dgusakov (changes requested):

tests/ui_tests/app/helpers/network.py
annademina: removed

tests/ui_tests/app/helpers/network.py
annademina: how long does it necessary to wait&

tests/ui_tests/app/helpers/network.py
annademina: added

Sealwing (approved):

Sealwing (approved):

dgusakov (approved):

PR #950, opened 2021-07-23
`config` field for `ConfigGroup` objects

dgusakov:


dgusakov:
Approves were received in #908

a-alferov (approved):

PR #953, opened 2021-07-29
ADCM-1928 Fix bug with not completely removed directory

SofiaTal:


dgusakov (approved):
I'm not sure about the exact code implementation. From the user side bug is fixed

PR #966, opened 2021-08-03
Update guest_version.sh

Maunty:
Script updated in order to provide branch name along with commit id

### DIFF
--- a/gues_version.sh
+++ b/gues_version.sh
@@ -14,7 +14,36 @@
 current_date=$(date '+%Y.%m.%d.%H')
 current_hash=$(git log --pretty=format:'%h' -n 1)
 
-printf '{\n\t"version": "%s",\n\t"commit_id": "%s"\n}\n' "${current_date}" "${current_hash}" > config.json
+name_rev=$(git name-rev --name-only HEAD)
+
+# posible outputs
+## local cloned
+# feature/ADH-1376/ADH-1377 #  branches
+# tags/0.1.4 #  tags
+## jenkins cloned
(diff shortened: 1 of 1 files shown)


Maunty:
tested with https://ci.arenadata.io/job/gpdb/job/adb-6.x/23/console.

gues_version.sh
```diff
@@ -14,7 +14,36 @@
...
+  origin/pr/* )
+    # we dont build adcm pr, only branches, just in case
+    current_branch="_$(echo "$name_rev" | cut -f 2,3 -d '/' | tr '/' '-')"
+    ;;
+  *)
+    current_branch="_$name_rev"
+    ;;
+esac
+
+printf '{\n\t"version": "%s",\n\t"commit_id": "%s"\n}\n' "${current_date}" "${current_hash}${current_branch}" > config.json
```
shvepsy, line 46:
```suggestion
printf '{\n\t"version": "%s",\n\t"commit_id": "%s",\n\t"branch": "%s"\n}\n' "${current_date}" "${current_hash}" "${current_branch}" > config.json
```

shvepsy: Can drop it if config.json isn't support `branch` field

shvepsy (approved):

gues_version.sh
Maunty: I think that we need @acmnu decision how is better to do. Currently i don't know what is frontend algorithm to gather adcm version from config,  so i decided to hack commit_id value.

PR #967, opened 2021-08-04
Adcm 1844_5 add ui tests for cluster page

annademina:


tests/ui_tests/app/page/cluster/page.py
```diff
@@ -110,6 +115,19 @@ def check_all_elements(self):
             ]
         )
 
+    def click_add_service_btn(self):
+        self.find_and_click(ClusterServicesLocators.add_services_btn)
+
+    @allure.step("Check service {service_name} in cluster")
+    def add_service_by_name(self, service_name: str):
```
Sealwing, line 124: The step is 'Check', but the method is 'add'
Looks more like "Add service {service_name} to cluster"

tests/ui_tests/app/page/cluster/page.py
```diff
@@ -54,35 +57,36 @@ def __init__(self, driver, base_url, cluster_id: int):
...
+        self.find_and_click(ObjectPageMenuLocators.actions_tab)
 
 
 class ClusterMainPage(ClusterPageMixin):
     """Cluster page Main menu"""
 
     MENU_SUFFIX = 'main'
 
+    @allure.step("Check main elements on page")
     def check_all_elements(self):
```
Sealwing, line 92: I suggest name allure step like "Check all main elements on the page are presented", so we know what exactly is checked
And there is another method with the same name, so maybe it's better to move it to Mixin and add an attribute like `MAIN_ELEMENTS` or `BASE_ELEMENTS`, so we can just override it in descendants

Sealwing (changes requested):

skhomuti (dismissed):

tests/ui_tests/app/page/cluster/page.py
annademina: yep, you're right, changed

tests/ui_tests/app/page/cluster/page.py
annademina: good idea! changed

Sealwing (approved):

skhomuti (approved):

acmnu:
bender build

BenderArenadata:

PR #972, opened 2021-08-05
ADCM-1937 fix bug with job_list

SofiaTal:
Need advice about correct behaviour of job_list.


python/cm/migrations/0068_alter_tasklog_action.py
```diff
@@ -0,0 +1,36 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Generated by Django 3.2.4 on 2021-08-05 14:31
+# pylint: disable=line-too-long
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
acmnu, line 19: @SofiaTal What about existing database with incorrect value of 'action' field, or this issue was not released?

acmnu (changes requested):

python/cm/migrations/0068_alter_tasklog_action.py
SofiaTal: I think that adding migration solve this problem, isn't it?

python/cm/migrations/0068_alter_tasklog_action.py
```diff
@@ -15,6 +15,15 @@
 from django.db import migrations, models
 import django.db.models.deletion
 
+def fix_joblog(apps, schema_editor):
+    JobLog = apps.get_model('cm', 'JobLog')
+    for job in JobLog.objects.all():
```
nagarin: should be single query, something like `JobLog.objects.filter(task=None).delete()`

nagarin (changes requested):

python/cm/migrations/0068_alter_tasklog_action.py
SofiaTal: It is done in last commit before your review

nagarin (approved):

nagarin:
https://stackoverflow.com/questions/38641952/django-filter-a-foreignkey-field-when-it-is-null

python/cm/migrations/0068_alter_tasklog_action.py
```diff
@@ -0,0 +1,45 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def fix_joblog(apps, schema_editor):
+    JobLog = apps.get_model('cm', 'JobLog')
+    LogStorage = apps.get_model('cm', 'LogStorage')
+    for job in JobLog.objects.filter(task_id__isnull=True):
+        LogStorage.objects.filter(job_id=job.id).delete()
```
acmnu, line 23:
```suggestion
        LogStorage.objects.filter(job__task_id__isnull=True).delete()
```

acmnu (changes requested):

nagarin (approved):

a-alferov (approved):

acmnu (approved):

PR #975, opened 2021-08-06
Added one url on list host for group config

a-alferov:


python/api/group_config/serializers.py
```diff
@@ -61,3 +56,15 @@ class Meta:
...
+        url = reverse(
+            viewname='host',
+            request=self.context['request'],
+            format=self.context['format'],
+        )
+        ids = [h.id for h in obj.hosts.all()]
+        view = self.context['view']
+        limit = view.paginator.limit
+        offset = view.paginator.offset
+        return f'{url}?ids={",".join(map(str, ids))}&limit={limit}&offset={offset}'
```
nagarin, line 70: May be someday we should replace such string manipulations with URL object manipulation like `furl` or `URLobject` or smth else

nagarin (approved):

nagarin (approved):

PR #983, opened 2021-08-11
ADCM-1926 Group configs

sanshan:


sanshan:
**What is not ready now**
I. First priority
- removing a host from a group
- groups config for services
- groups config for components

I. Second priority
- events for groups config
- history for groups config
- available host list for each group

drKaramazin (approved):

PR #997, opened 2021-08-16
ADCM-1850 UI tests for /profile

Sealwing:


tests/ui_tests/test_profile_page.py
```diff
@@ -0,0 +1,85 @@
...
+        intro_page = AdminIntroPage(app_fs.driver, app_fs.adcm.url)
+        intro_page.wait_config_loaded()
+        intro_page.header.open_profile()
+        profile_page.wait_page_is_opened()
+        profile_page.check_required_fields_are_presented()
+        _assert_username_on_profile_page(profile_page, adcm_credentials['username'])
+
+
+@allure.step('Check username on profile page is {expected_username}')
+def _assert_username_on_profile_page(page: ProfilePage, expected_username: str):
```
vgorkavenko, line 77: I think this and similar steps should be in `*Page` classes. Like `check_required_fields_are_presented`

tests/ui_tests/app/api.py
```diff
@@ -0,0 +1,82 @@
...
+        caller_function_name = inspect.stack()[1][3].replace('_', ' ')
+        try:
+            response_content = response.json()
+        except json.JSONDecodeError:
+            response_content = response.text
+        raise RequestFailedException(
+            f'Failed to {caller_function_name} over API. '
+            f'Request finished with status {status_code} '
+            f'and json body: {response_content}',
+        )
```
vgorkavenko, line 82: There is no need to write func name in the error since it will still occur inside some allure step

tests/ui_tests/test_profile_page.py
```diff
@@ -0,0 +1,85 @@
...
+    profile_page.check_required_fields_are_presented()
+    _assert_username_on_profile_page(profile_page, another_user['username'])
+
+
+@pytest.mark.usefixtures("login_to_adcm_over_api")
+def test_change_password(adcm_credentials: dict, app_fs: ADCMTest):
+    """Change admin password over UI and login under new credentials"""
+    new_password = 'password'
+    profile_page = ProfilePage(app_fs.driver, app_fs.adcm.url).open()
+    with restore_admin_password(new_password, adcm_credentials, app_fs.adcm.url) as new_credentials:
```
vgorkavenko, line 63: I think there is no need to change admin password back. This is a test of changing the password from the default to a new one and it shouldn't contain any additional hidden steps

tests/ui_tests/test_profile_page.py
Sealwing: Good point

tests/ui_tests/app/api.py
Sealwing: Thought it would be nice to add a name to the exception itself, but on other hand, it's not very useful without a report. Removed.

tests/ui_tests/test_profile_page.py
Sealwing: @vgorkavenko During teardown ADCM fixture removes hosts. That action requires authentication. As we changed the admin password in test, we need to set it back. I guess, it'll be better to make a fixture that sets the password back again, but not sure about order of teardown calls.
Maybe smt like code below will work?
```python
@pytest.fixture(params=[{'username': 'admin', 'password': 'newpass'}])
def restore_password(request, adcm_credentials, adcm_fs):
    new_creds = request.param
    ...
```

Sealwing: Reworked as mentioned above

vgorkavenko: Do we create real hosts in some UI tests?

Sealwing: Nope, its just default implementation of a teardown in `_adcm` method that is used in `adcm_*` fixtures

vgorkavenko: I think it needs not to create a separate fixture because of this and somehow handle this situation in tests in which we do not create real hosts. @dgusakov ?

tests/ui_tests/test_cluster_list_page.py
```diff
@@ -508,9 +508,10 @@ def test_check_pagination_on_cluster_host_page(
     ):
         cluster = create_community_cluster
         provider = upload_and_create_provider
-        for i in range(11):
-            host = provider.host_create(f"{HOST_NAME}_{i}")
-            cluster.host_add(host)
+        with allure.step('Create 11 hosts'):
+            for i in range(11):
```
dgusakov, line 512:
```suggestion
        hosts_count = 11
        with allure.step(f'Create {hosts_count} hosts'):
            for i in range(hosts_count):
```

tests/ui_tests/app/api.py
```diff
@@ -0,0 +1,79 @@
...
+import json
+import allure
+import requests
+
+
+class RequestFailedException(Exception):
+    """Request to ADCM API has status code >= 400"""
+
+
+class APIRequester:
```
dgusakov, line 21: Consider renaming this class to something more elegant. Like ADCMApiClient

tests/ui_tests/test_profile_page.py
dgusakov: @Sealwing This call https://github.com/arenadata/adcm_pytest_plugin/blob/a2b5595e1af39c2384c0628d2b985b83e6e9d5fe/src/adcm_pytest_plugin/fixtures.py#L189 should be removed since there is other places where we should remove hosts. After step removal from adcm_pytest_plugin credentials restores will become useless here.

tests/ui_tests/test_cluster_list_page.py
Sealwing: Sure

tests/ui_tests/app/api.py
Sealwing: Good point, renamed to "ADCMDirectAPIClient" to distinct it from "ADCMClient" and keep ADCM and API terms cases readable

dgusakov (dismissed):

dgusakov:
bender build

BenderArenadata:
Hey, @dgusakov! Build has been launched!

annademina (approved):

Sealwing:
bender build

BenderArenadata:
Hey, @Sealwing! Build has been launched!

Sealwing:
Bender build

BenderArenadata:
Hey, @Sealwing! Build has been launched!

Sealwing:
Bender build

BenderArenadata:
Hey, @Sealwing! Build has been launched!

PR #998, opened 2021-08-17
ADCM-1968 Close all tabs in test teardown before container is stopped

skhomuti:


tests/ui_tests/conftest.py
```diff
@@ -172,12 +171,14 @@ def _write_json_file(f_name, j_data):
 def login_to_adcm_over_api(app_fs, adcm_credentials):
     """Perform login via API call"""
     login_endpoint = f'{app_fs.adcm.url.rstrip("/")}/api/v1/token/'
-    app_fs.driver.get(app_fs.adcm.url)
+    LoginPage(app_fs.driver, app_fs.adcm.url).open().wait_page_is_opened()
```
Sealwing, line 174: You don't need to explicitly wait for the page to be opened, it's implicit in open(). `wait_page_is_opened` exists for cases, when we open page by click on some UI element, for example

skhomuti: fixed

tests/ui_tests/conftest.py
```diff
@@ -67,13 +67,6 @@ def app_fs(adcm_fs: ADCM, web_driver: ADCMTest, request):
     Attach ADCM API to ADCMTest object and open new tab in browser for test
     Collect logs on failure and close browser tab after test is done
     """
-    try:
-        web_driver.new_tab()
-    # Recreate session on WebDriverException
-    except WebDriverException:
-        # this exception could be raised in case
-        # when all tabs were closed in process of creating new one
-        web_driver.create_driver()
```
dgusakov, line 76: This route was meant to handle situations of unexpected driver failures. How it will be handled now?

dgusakov (changes requested):

tests/ui_tests/conftest.py
skhomuti: The comment suggests that all tabs were closed, but now it is an impossible case
I have not seen unexpected driver failures. Do you have examples?

Sealwing (approved):

tests/ui_tests/conftest.py
dgusakov: Nope. But additional insurance is useful)

Sealwing (approved):

Sealwing (approved):

dgusakov (approved):

PR #999, opened 2021-08-17
ADCM-1922: issue as concern

nagarin:


python/cm/models.py
```diff
@@ -211,32 +221,65 @@ class ADCMEntity(ADCMModel):
...
+        """Add issue to its source"""
+        for _ in issue.source_objects:
+            raise AdcmEx('ISSUE_INTEGRITY_ERROR')
+
+        self._issues.add(issue)
+
+    def get_issue(self, issue_type: IssueType) -> Optional['ConcernItem']:
+        """Get object's issue of specified type or None"""
+        issue_name = self.gen_issue_name(issue_type)
+        for issue in self._issues.filter(name=issue_name):
```
grey-kristy, line 277: Are we expected exactly one issue here? Or return only the first one?

python/cm/models.py
```diff
@@ -211,32 +221,65 @@ class ADCMEntity(ADCMModel):
     config = models.OneToOneField(ObjectConfig, on_delete=models.CASCADE, null=True)
     state = models.CharField(max_length=64, default='created')
     _multi_state = models.JSONField(default=dict, db_column='multi_state')
-    issue = models.JSONField(default=dict)
-    concern = models.ManyToManyField('ConcernItem', blank=True, related_name='%(class)s_entities')
+    _issues = models.ManyToManyField(
+        'ConcernItem', blank=True, db_column='issues', related_name='%(class)s_sources'
```
grey-kristy, line 225: Why we use db_column here, not just field name without _?

python/cm/issue.py
```diff
@@ -10,135 +10,44 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import defaultdict
-
-import cm.status_api
 from cm.adcm_config import proto_ref, obj_ref, get_prototype_config
 from cm.errors import AdcmEx, raise_AdcmEx as err
 from cm.hierarchy import Tree
```
grey-kristy, line 15: I think Tree is too generic name for this class. Maybe something more specific like ClusterTree?

grey-kristy (commented):
Does not we have problems with double lock/unlock the same objects from different tasks?

python/cm/models.py
nagarin: one or zero, more than one of same type (and same name) is imposible because of constraint

python/cm/models.py
nagarin: underscore is used to hide it, of course. it's kind of internals not needed in API

nagarin:
bender build

BenderArenadata:
Hey, @nagarin! Build has been launched!

nagarin:
bender build

BenderArenadata:
Hey, @nagarin! Build has been launched!

PR #1064, opened 2021-09-07
ADCM-1987 add graph realtions to Role model

grey-kristy:


python/rbac/upgrade_role.py
```diff
@@ -0,0 +1,150 @@
...
+        new_role.save()
+    if 'description' in role:
+        new_role.description = role['description']
+    for perm in perm_list:
+        new_role.permissions.add(perm)
+    new_role.save()
+    return new_role
+
+
+def get_role_spec():
```
acmnu, line 109: I don't like this. This function is not testable for no reason. I think it is better to add parameter "role_filename" or something like this and apply a constant in call time.

acmnu (changes requested):
Please modify code with the following logic:
* Add docstring to every class and function
* Add type hinting in every place you expect the exact type.

python/rbac/models.py
```diff
@@ -0,0 +1,128 @@
...
+        group_roles = group.role_set.all()
+        if self not in group_roles:
+            err('ROLE_ERROR', f'Group "{group.name}" does not has role "{self.name}"')
+        perm_list = self.get_permissions_without_role(group_roles)
+        with transaction.atomic():
+            self.group.remove(group)
+            self.save()
+            for perm in self.get_permissions():
+                if perm.codename not in perm_list:
+                    group.permissions.remove(perm)
```
kuhella, line 114: This should be part of Policy not a Role.

PR #1070, opened 2021-09-09
experiment about changing AUTH_USER_MODEL in the middle of the project

grey-kristy:


acmnu:
There is no needs in this anymore

PR #1082, opened 2021-09-14
ADCM-2051 HTTP 400 in case of wrong object_id

acmnu:
This commit changes returned HTTTP code for POST on group-config
when object_id is not valid (object does not exists)


nagarin (dismissed):

acmnu:
bender build

BenderArenadata:

PR #1093, opened 2021-09-17
Fixed f-strings and logging interpolation pylint rules

skhomuti:


Sealwing (approved):

python/cm/api.py
```diff
@@ -73,7 +73,7 @@ def add_cluster(proto, name, desc=''):
         cm.issue.update_hierarchy_issues(cluster)
     cm.status_api.post_event('create', 'cluster', cluster.id)
     cm.status_api.load_service_map()
-    log.info(f'cluster #{cluster.id} {cluster.name} is added')
+    log.info('cluster #%s %s is added', cluster.id, cluster.name)
```
vgorkavenko, line 76: Why not `f` string?

skhomuti: https://docs.python.org/3/howto/logging.html#optimization

vgorkavenko (approved):

madlabman (approved):

Sealwing (approved):

PR #1103, opened 2021-09-21
Adcm-2083 add tests for deleting host from group

annademina:


tests/functional/test_config_groups.py
```diff
@@ -79,6 +79,22 @@ def create_two_hosts(provider) -> Tuple[Host, Host]:
     return test_host_1, test_host_2
 
 
+@pytest.fixture()
+def cluster_with_components_on_one_host(
```
Sealwing, line 83: Suggestion: since you have `FIRST_HOST` constant and here you bind both components on "first" host, so name this function to `cluster_with_components_on_one_host`.
The main profit is that when one's reading code where this fixture is used, it's clear that both components are bind to the "first" host, not "one of the hosts".

tests/functional/test_config_groups.py
```diff
@@ -260,3 +261,69 @@ def test_add_incorrect_host_to_component_group(self, cluster_with_components_on_
...
+        assert len(group.hosts()) == 0, "Should not be any hosts in conf group"
+
+    def test_delete_host_from_group_after_deleting_in_cluster(self, sdk_client_fs, cluster, provider):
+        """Test that host removed from conf group after removing from cluster"""
+
+        test_host = provider.host_create(fqdn=FIRST_HOST)
+        cluster.host_add(test_host)
+        with allure.step("Create config group for cluster and add the host"):
+            cluster_group = cluster.group_config_create(name=FIRST_GROUP)
+            cluster_group.host_add(test_host)
```
Sealwing, line 280: Maybe add a check that host is actually in group (maybe even make method-step) here?
If it isn't checked we won't know that deletion worked.

Sealwing (commented):
Isn't there any method to remove the host from the group without removing it from the hostcomponent? e.g. when I added the wrong host and want to remove it from group. Or is it tested somewhere else?

tests/functional/test_config_groups.py
annademina: renamed to cluster_with_components_on_first_host

tests/functional/test_config_groups.py
annademina: added checks

Sealwing (approved):

skhomuti (approved):

PR #1110, opened 2021-09-24
ADCM-2089 pin jinja and markup versions

skhomuti:


dgusakov (approved):

dgusakov:
@skhomuti Pls leave a note about why we fix these versions

dgusakov (approved):

PR #1111, opened 2021-09-27
ADCM-2042 Added status changer for using in tests

Sealwing:


tests/library/status.py
```diff
@@ -0,0 +1,113 @@
...
+
+    If you want to "enable" Cluster or Service,
+        you should set positive status on all hosts and components on it.
+    If you want to "disable" Cluster or Service,
+        you should set negative status at least one of its components or hosts.
+
+    After using status-changing methods you'll need to call `reread` method
+        to be able to get newly set status.
+
+    Consider that status != 16 will be reset to 16 after timeout (300 seconds).
```
madlabman, line 38: Status value 16 has a specific meaning?

Sealwing: Not totally sure, but I think it's "default" status when no info from statuschecker was received

Sealwing: if component wasn't found in db map 16 will be returned
``` go
func (db dbMap2) get(key1 int, key2 int) (Status, bool) {
	val, ok := db[key1][key2]
	if ok {
		return val, true
	}
	return Status{Status: 16}, false
}
```

tests/library/status.py
```diff
@@ -0,0 +1,113 @@
...
+            if host_id not in hosts:
+                self._set_host_status(host_id, POSITIVE_STATUS)
+                hosts.add(host_id)
+
+    def set_host_positive_status(self, hosts: Collection[Host]) -> None:
+        """Make hosts "working" by setting status to 0"""
+        for host in hosts:
+            self._set_host_status(host.id, POSITIVE_STATUS)
+
+    def set_host_negative_status(self, hosts: Collection[Host], status: int = 16) -> None:
```
skhomuti, line 66: `host` or `hosts`?

tests/library/status.py
```diff
@@ -0,0 +1,113 @@
...
+            self._set_status(self.host_url_template.format(host.id, component.id), POSITIVE_STATUS)
+
+    def set_component_negative_status(
+        self, host_components: Collection[Tuple[Host, Component]], status: int = 16
+    ) -> None:
+        """Make components "not working" by settings status != 0"""
+        for host, component in host_components:
+            self._set_status(self.host_url_template.format(host.id, component.id), status)
+
+    def _set_host_status(self, host_id: int, status: int) -> None:
```
skhomuti, line 83: this function will be useful in public scope with `Host` argument type, not id
same for component

skhomuti (changes requested):

tests/library/status.py
Sealwing: @skhomuti 
Do you mean it'll be useful to be able to pass a single host instead of a collection?
This particular function is made to be used in both cases: when we do not have a host/component objects and when we do.
I can make set status functions to take either collection or a single value, what do you think?

tests/library/status.py
Sealwing: yeah, hosts, actually, but check out discussion below

tests/library/status.py
skhomuti: yep, collection or a single value in arguments will be ok

Sealwing: then naming will be "set_host_*" and "set_component_*" I think

skhomuti (approved):

madlabman (dismissed):

skhomuti (approved):

dgusakov (approved):

PR #1113, opened 2021-09-27
Fix ADCM schema for action masking

nagarin:


python/cm/models.py
```diff
@@ -876,6 +876,12 @@ def get_id_chain(self, target_ids: dict) -> dict:
 
     def allowed(self, obj: ADCMEntity) -> bool:
         """Check if action is allowed to be run on object"""
+        if self.state_unavailable == 'any':
+            return False
+
+        if self.multi_state_unavailable == 'any':
+            return False
+
```
dgusakov, line 884: So we drop support of ANy here?
@acmnu Are you ok with it?

Sealwing (approved):

python/cm/models.py
nagarin: Otherwise, I added support for 'any' in unavailable, as it was discussed.

dgusakov: Ok. Thx for clarifying

acmnu:
@Sealwing Is it done?

Sealwing (approved):

vgorkavenko (approved):

PR #1118, opened 2021-09-28
ADCM-2066 Add display names to services and components

SofiaTal:


nagarin (commented):
look at ADCMEntity's same-named property - it covers these cases

nagarin (approved):

grey-kristy (approved):

PR #1123, opened 2021-09-29
ADCM-2092 Bump ansible to 2.9.25

acmnu:


dgusakov:
We need to run a full set of bundle tests on this branch. But until develop branch is red that will make no sense.

PR #1143, opened 2021-10-05
ADCM-2137 Added creating file for file type field

a-alferov:


acmnu:
@a-alferov we have red functional tests still.

PR #1147, opened 2021-10-06
Adcm-2086 add negative tests to config groups

annademina:


tests/functional/test_config_groups.py
```diff
@@ -448,6 +418,24 @@ def _get_config_from_group(self, cluster_group: GroupConfig):
...
+            with pytest.raises(ErrorMessage) as e:
+                group.config_set_diff(params)
+            ATTRIBUTE_ERROR.equal(e)
+        with allure.step(f'Check error message is "{HOST_EXISTS_MESSAGE}"'):
+            assert (
+                ATTRIBUTE_ERROR_MESSAGE in e.value.error['desc']
+            ), f"Should be error message '{ATTRIBUTE_ERROR_MESSAGE}'"
+
+    def _check_error_in_all_params_in_group(self, group: GroupConfig, config_before: dict):
+        for param in config_before.keys():
```
skhomuti, line 433: add allure step for each param
now it is not obviously
image
upd: will be fixed in #1144

Sealwing (dismissed):

tests/functional/test_config_groups.py
```diff
@@ -362,69 +364,68 @@ class TestChangeGroupsConfig:
...
-                "option": option,
-                "text": text,
-                "group": group,
-                "structure": structure,
-                "map": map_value,
-            },
-        }
+    def _add_values_to_group_config_template(
+        self, custom_group_keys: dict = {}, group_keys: dict = {}, config_attr: dict = PARAMS_TO_CHANGE
+    ) -> dict:
```
Sealwing, line 409: Please provide a docstring and why default values are dicts and not None's?

tests/functional/test_config_groups.py
```diff
@@ -448,109 +449,348 @@ def _get_config_from_group(self, cluster_group: GroupConfig):
         config_group["file"] = config_group["file"].replace("\n", "")
         return config_group
 
+    @allure.step('Check error that group config cant change')
```
Sealwing, line 452: Maybe reverse `'` and `"` and use "can't"?

tests/functional/test_config_groups.py
```diff
@@ -362,69 +364,68 @@ class TestChangeGroupsConfig:
...
+                group_config_template["attr"]["custom_group_keys"]["group"] = {
+                    "port": group_value,
+                    "transport_port": group_value,
+                }
+            for key, value in custom_group_keys.items():
+                group_config_template["attr"]["custom_group_keys"][key] = value
+        if group_keys:
+            group_config_template["attr"]["group_keys"] = {}
+            for key, value in group_keys.items():
+                group_config_template["attr"]["group_keys"][key] = value
```
Sealwing, line 424: as I understand, you want to extract group_keys to `group_config_template["attr"]["group_keys"]`, so you can use `group_config_template["attr"]["group_keys"] = {**group_keys}` or `deepcopy`. Same for code above and below.

tests/functional/test_config_groups.py
```diff
@@ -362,69 +364,68 @@ class TestChangeGroupsConfig:
...
+    ) -> dict:
+        group_config_template = {"attr": {}, "config": {}}
+        if custom_group_keys:
+            group_config_template["attr"]["custom_group_keys"] = {}
+            if "group" in custom_group_keys.keys():
+                group_value = custom_group_keys.pop("group")
+                group_config_template["attr"]["custom_group_keys"]["group"] = {
+                    "port": group_value,
+                    "transport_port": group_value,
+                }
```
Sealwing, line 418: Why do you want to hardcode keys here instead of always passing group as `{"port": x, "transport_port": x}`?  Aren't there any cases, when you may want to pass something else in the "group" key? If it's required to be that way please consider mentioning this behavior in docstring or comment.

annademina: changed

tests/functional/test_config_groups.py
annademina: changed

tests/functional/test_config_groups.py
annademina: changed

skhomuti (approved):

PR #1164, opened 2021-10-08
ADCM-2139 add ansible_ssh_private_key_file

SofiaTal:


python/cm/management/commands/loadcluster.py
```diff
@@ -119,16 +120,35 @@ def create_provider(provider):
...
         return ex_id, provider
 
 
+def create_needed_file(obj, config):
+    conf = config["current"]["config"]
+    proto = obj.prototype
+    for key, value in conf.items():
+        if isinstance(value, dict):
+            for subkey, subvalue in value.items():
+                pconf = models.PrototypeConfig.objects.get(
```
nagarin, line 141: It may be look like premature optimization, but anyway:
`if models.PrototypeConfig.objects.filter(...., type='file').exists():
    save_file_type(...)
`

nagarin (changes requested):

python/cm/management/commands/loadcluster.py
SofiaTal: Done in next commit

nagarin (dismissed):

python/cm/management/commands/loadcluster.py
```diff
@@ -84,6 +85,23 @@ def create_config(config):
         return None
 
 
+def create_file_from_config(obj, config):
+    if config is not None:
+        conf = config["current"]["config"]
+        proto = obj.prototype
+        for key, value in conf.items():
+            if isinstance(value, dict):
```
a-alferov, line 93: We cannot trust this condition as we have types such as `structure` and `map`, I think you need to create files based on the specification, and take the values from the config

PR #1176, opened 2021-10-11
ADCM-2128: update Task lock message

nagarin:


python/cm/job.py
```diff
@@ -92,15 +93,13 @@ def check_action_hosts(action, obj, cluster, hosts):
...
-    action, obj, conf, attr, hc, hosts, event, verbose
-):  # pylint: disable=too-many-locals
+def prepare_task(action, obj, conf, attr, hc, hosts, verbose):  # pylint: disable=too-many-locals
     cluster = get_object_cluster(obj)
     check_action_state(action, obj, cluster)
     _, spec = check_action_config(action, obj, conf, attr)
     if conf and not spec:
         err("CONFIG_VALUE_ERROR", "Absent config in action prototype")
-    host_map, delta = check_hostcomponentmap(cluster, action, hc)
+    host_map, _ = check_hostcomponentmap(cluster, action, hc)
```
grey-kristy, line 102: It seems nobody expect delta from check_hostcomponentmap() so we should remove if from check_hostcomponentmap() return values

python/cm/job.py
```diff
@@ -532,31 +529,33 @@ def create_one_job_task(action, obj, conf, attr, hc, hosts, event, verbose):
         finish_date=timezone.now(),
         status=config.Job.CREATED,
     )
-    task.save()
+    set_task_status(task, config.Job.CREATED, ctx.event)
+
+    if action.type == ActionType.Job.value:
+        sub_actions = [None]
```
grey-kristy, line 535: Maybe I'm wrong, but it seems in case of one job task you don't create any job?

grey-kristy (dismissed):

PR #1197, opened 2021-10-18
ADCM-2169 add job_lock() to all ansible plugins

grey-kristy:


python/cm/ansible_plugin.py
```diff
@@ -208,6 +234,7 @@ def run(self, tmp=None, task_vars=None):  # pylint: disable=too-many-branches
             raise AnsibleError(MSG_NO_ROUTE)
 
         result = super().run(tmp, task_vars)
+        job_unlock(lock)
```
acmnu, line 237: I see no unlock in case of raise in line 234. Is this a trouble?

PR #1206, opened 2021-10-20
Fix flaky tests: `test_custom_name_config` and `test_invoker_object_url`

Sealwing:
Fix was made by changing the logic of underlying functions working with UI data and inputs

### DIFF
--- a/tests/ui_tests/utils.py
+++ b/tests/ui_tests/utils.py
@@ -26,6 +26,7 @@ from adcm_pytest_plugin.utils import random_string, wait_until_step_succeeds
 from selenium.webdriver.remote.webdriver import WebDriver
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait as WDW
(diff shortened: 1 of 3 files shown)


tests/ui_tests/utils.py
```diff
@@ -167,6 +168,21 @@ def check_rows_amount(page, expected_amount: int, table_page_num: int):
...
+    """
+    Use it as decorator to catch flaky exceptions and raise Assertion messages instead.
+    Use it only with `wait_until_step_succeeds` or something similar.
+    """
+
+    def wrapped(*args, **kwargs):
+        try:
+            func(*args, **kwargs)
+        except (StaleElementReferenceException, NoSuchElementException) as e:
+            raise AssertionError('Got a flaky error: %s' % e) from e
```
madlabman, line 181: I guess you should use f-string here.

Sealwing: yep, % is preferred for error message formatting, but here it serves no purpose. Changed

madlabman (approved):

PR #1231, opened 2021-10-29
ADCM-2250 Fix expected error codes

Sealwing:


Sealwing:
@skhomuti Please verify the last change of the `__main_info` test. Not sure it doesn't break test logic, but otherwise I get "new" error

skhomuti (approved):

PR #1238, opened 2021-11-02
QA-315 Up Selenium version to 4.0.0

Sealwing:


dgusakov:
@Sealwing There is no need to open PR if you just need to run tests on CI.

PR #1251, opened 2021-11-09
ADCM-2245 Added `rbac/policy` endpoint

a-alferov:


python/rbac/endpoints/policy/services.py
```diff
@@ -0,0 +1,151 @@
...
+    users = kwargs.get('user', ())
+    policy.user.add(*users)
+
+    groups = kwargs.get('group', ())
+    policy.group.add(*groups)
+
+    return policy
+
+
+def update_m2m_field(m2m, instances) -> None:
```
nagarin, line 95: type hints missed

python/rbac/endpoints/policy/services.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    Update m2m field for object
+
+    :param m2m: ManyToManeField
+    :type m2m: ManyRelatedManager
+    :param instances: list of objects
+    :type instances: list
+    """
+    if instances:
```
nagarin, line 104: m2m.clear()
if instances:
    m2m.add...

nagarin (changes requested):

python/rbac/endpoints/policy/services.py
```diff
@@ -0,0 +1,151 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
nagarin, line 1: I think, general idea was to have _services_ as level completely separated from models and from endpoints as well - in `python/rbac/services/` directory for example.

PR #1253, opened 2021-11-10
ADCM-2248 Cleanup and new test cases for import tests

Sealwing:


dgusakov (dismissed):

dgusakov (approved):

tests/functional/test_import.py
```diff
@@ -94,3 +98,22 @@ def test_cluster_import(sdk_client_fs: ADCMClient, path):
         cluster_import = bundle_import.cluster_create("cluster_import")
     with allure.step('Bind cluster from cluster with export to cluster with import'):
         cluster_import.bind(cluster)
+
+
+@pytest.mark.xfail(reason='https://arenadata.atlassian.net/browse/ADCM-2292')
+def test_import_with_zero_range(sdk_client_fs: ADCMClient):
```
alexterent, line 104: Mb call it "test_cluster_import_with_zero_range"? Or bundle imported too? Then mb docstring need to include "cluster and bundle".

annademina (approved):

tests/functional/test_import.py
Sealwing: It's not important what we bind here, bug can be reproduced both on cluster and import, I guess mention of a cluster in docstring is enough

alexterent (approved):

PR #1287, opened 2021-11-22
ADCM-2335 Fix default file value to ui_config

SofiaTal:


python/cm/adcm_config.py
```diff
@@ -72,6 +72,8 @@ def to_flat_dict(conf, spec):
 
 def get_default(c, proto=None):  # pylint: disable=too-many-branches
     value = c.default
+    if proto is None:
+        proto = Prototype.objects.get(id=c.prototype_id)
```
a-alferov, line 76: I think this code is superfluous here. We only need a prototype to find out the hash of the bundle and then build the path to the origin file.

SofiaTal: Not only for building path. In read_file_type func we use proto and proto.bundle.hash. 
In some places of code we use get_default without passing prototype and it leads to trouble with reading file type. I could add proto in the exact place in ui_config but the error may be reproduced in other place. So i think we should fix it here

a-alferov (approved):

PR #1296, opened 2021-11-24
ADCM-2356 Add rbac user endpoint

skhomuti:
Add rbac group draft endpoint
Fix patch data preparing on methods. Now PATCH body contains only changeable fields instead of full body

### DIFF
--- a/tests/api/utils/types.py
+++ b/tests/api/utils/types.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 """Module contains all field types and it special values"""
+import random
 import warnings
 from abc import ABC, abstractmethod
 from collections.abc import Callable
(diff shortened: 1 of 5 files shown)


tests/api/utils/types.py
```diff
@@ -251,6 +263,17 @@ def generate(self, **kwargs):
...
 
+class Email(BaseType):
+    """Email field type"""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._sp_vals_negative = [PreparedFieldValue(f_type=Type[String]), PreparedFieldValue(f_type=Type[PositiveInt])]
+
+    def generate(self, **kwargs):
+        return f"{random_string(10)}@{random_string(5)}.com"
```
Sealwing, line 274: Only `.com` is supported?

Sealwing (approved):

tests/api/utils/types.py
skhomuti: nope, it's just a simple valid value

PR #1301, opened 2021-11-25
ADCM-2309 Add parametrized block of roles

SofiaTal:
Without role "Assign roles" and "View status information"


python/rbac/upgrade/role_spec.yaml
```diff
@@ -59,6 +56,737 @@ roles:
...
+      module_name: rbac.roles
+      class_name: ObjectRole
+      init_params:
+        app_name: cm
+        model: ADCM
+        filter:
+          name: ADCM
+      apps:
+        - label: cm
+          models:
```
nagarin, line 596: Is it different from another "compare * settings"? it does not include objectconfig model

PR #1302, opened 2021-11-25
Fixed filter event for new adwp-evens version

a-alferov:


nagarin (approved):

acmnu (changes requested):
Why?

a-alferov:
Doesn't make sense

PR #1317, opened 2021-12-01
ADCM-2352 add aggregate roles for actions

grey-kristy:


python/cm/bundle.py
```diff
@@ -249,13 +249,16 @@ def update_obj(dest, source, fields):
...
 
 def cook_roles(bundle):
+    parent = {}
     for act in Action.objects.filter(prototype__bundle=bundle):
         name = act.display_name
         model = get_model_by_type(act.prototype.type)
         role = Role(
             name=f'{name}_{act.prototype.type}_{act.prototype.display_name}',
             description=f'run action {name} of {act.prototype.type} {act.prototype.display_name}',
+            category=f'{bundle.name}',
```
acmnu, line 259: * category is a list, doesn't it?
* Use a "display_name" of bundle, not a name.

acmnu (changes requested):

python/cm/bundle.py
grey-kristy: bundle does not has a display_name

acmnu: @grey-kristy It can't be true. Bundle have no its own properties, so it takes definition of cluster or provider and both of them has "display_name".

acmnu (approved):

PR #1323, opened 2021-12-02
ADCM-2343  and ADCM-2372 API tests on /rbac/role and /rbac/policy

Sealwing:


tests/api/testdata/db_filler.py
```diff
@@ -194,11 +197,29 @@ def _solve_field_relations(self, endpoint: Endpoints, data: dict, field: Field,
...
         else:
             raise NotImplementedError(f"Relations logic needs to be implemented for {endpoint} for field {field.name}")
 
         return _data, field
 
+    def _get_adcm_object_id_by_object_type(
+        self, object_type: Literal["cluster", "service", "component", "provider", "host"]
+    ) -> int:
+        """Get random created object by given type"""
+        return random.choice(get_endpoint_data(adcm=self.adcm, endpoint=Endpoints[object_type.capitalize()]))["id"]
```
skhomuti, line 221: why you do not use enum in the field `object_type`?

tests/api/testdata/db_filler.py
```diff
@@ -248,9 +269,9 @@ def _choose_fk_field_value(self, field: Field, fk_data: list):
         """Choose a random fk value for the specified field"""
         if isinstance(field.f_type, ForeignKey):
             fk_class_name = field.f_type.fk_link.__name__
-            if fk_class_name in self._available_fkeys:
+            if available_fkeys := self._available_fkeys.get(fk_class_name, None):
```
skhomuti, line 272: please add a comment to explain why was it required

skhomuti: `if fk_vals := ...`

tests/api/testdata/generators.py
```diff
@@ -87,6 +86,8 @@ def _fill_pytest_param(
     if endpoint.spec_link:
         marks.append(allure.link(url=endpoint.spec_link, name="Endpoint spec"))
     param_id = f"{endpoint.path.replace('/', '_')}_{method.name}_{positive_str}"
+    if sum(1 for e in Endpoints if e.path == endpoint.path) > 1:
+        param_id += f"_{endpoint.endpoint.get_data_class_pretty_name()}"
```
skhomuti, line 90: probably you can move it to the endpoint, like something `endpoint.pretty_name()`
return path by default or path + dataclass if it's needed

tests/api/testdata/getters.py
```diff
@@ -29,11 +29,15 @@ def get_endpoint_data(adcm: ADCMTestApiWrapper, endpoint: Endpoints) -> list:
...
-        return res.json()
+    result = res.json()
     # New endpoints always return a response with pagination.
     # In the future all endpoints will return that
-    return res.json().get("results")
+    if not isinstance(result, list):
+        result = result.get("results")
+    if endpoint.endpoint.filter_predicate is None:
+        return result
+    # Is used mostly for "special" or "corner-case" endpoints
```
skhomuti, line 39: not really. it is used for endpoints with more than one dataclass
"special or corner-case" does not explain

tests/api/utils/types.py
```diff
@@ -406,6 +419,41 @@ def __init__(self, **kwargs):
         ]
 
 
+class GenericForeignKeyList(BaseType):
```
skhomuti, line 422: seems like you can use `return_value` func

skhomuti (changes requested):

tests/api/utils/types.py
Sealwing: I guess it's kind of parallel thing, more of a wrapper, not a data receiver, so not sure how to use it instead of `payload` workaround

tests/api/testdata/getters.py
Sealwing: Changed

tests/api/testdata/generators.py
Sealwing: sure

tests/api/testdata/db_filler.py
Sealwing: Tried to add decent explanation here

tests/api/testdata/db_filler.py
Sealwing: not the case

tests/api/testdata/generators.py
```diff
@@ -87,7 +87,7 @@ def _fill_pytest_param(
         marks.append(allure.link(url=endpoint.spec_link, name="Endpoint spec"))
     param_id = f"{endpoint.path.replace('/', '_')}_{method.name}_{positive_str}"
     if sum(1 for e in Endpoints if e.path == endpoint.path) > 1:
-        param_id += f"_{endpoint.endpoint.get_data_class_pretty_name()}"
+        param_id += f"_{endpoint.pretty_name}"
```
skhomuti, line 90: two lines above can be in pretty_name realization too

Sealwing: discussed, reworked

tests/api/testdata/db_filler.py
```diff
@@ -194,11 +202,29 @@ def _solve_field_relations(self, endpoint: Endpoints, data: dict, field: Field,
                 ][-1]
                 field.f_type.schema = build_schema_by_json(current_config_log[field.name])
 
+        elif endpoint == Endpoints.RbacNotBuiltInPolicy:
+            if field.name == "object":
+                role_fk = _data[related_field_name]
+                role = get_object_data(adcm=self.adcm, endpoint=endpoint.RbacAnyRole, object_id=role_fk)
+                _data[field.name] = field.f_type.payload = [
```
skhomuti, line 209: payload is still needed?

skhomuti (changes requested):

tests/api/testdata/db_filler.py
Sealwing: Yep, I use it in other places. I guess we need to cover more than one place for such "special" stuff anyway

Sealwing: I've reworked `payload` a little, made some fixes, so it'll be called directly in `_prepare_field_value_for_object_creation`

vgorkavenko (approved):

skhomuti (approved):

PR #1328, opened 2021-12-02
ADCM-2401 Change business_perm to enum type for role

SofiaTal:


python/cm/bundle.py
```diff
@@ -258,7 +258,7 @@ def cook_roles(bundle):
             description=f'run action {name} of {act.prototype.type} {act.prototype.display_name}',
             category=[f'{bundle.name}'],
             bundle=bundle,
-            business_permit=False,
+            type='role',
```
nagarin, line 261: That's what I talked about - this should be like that:
`type = rbac.model.RoleTypes.role`
in order to not mistake or typo or to be able to find/rename every use of that constant

python/rbac/services/role.py
```diff
@@ -23,18 +23,16 @@ def check_role_child(child: List[Role]) -> None:
         if not item.built_in:
             errors = {'child': ['Only built-in roles allowed to be included as children.']}
             raise ValidationError(errors)
-        if item.child.all():
-            errors = {
-                'child': ['Only 2 levels allowed, so you can’t connect to role who has a child.']
-            }
+        if item.type != 'business':
```
nagarin, line 26: here is the same as before

nagarin (changes requested):

python/cm/bundle.py
SofiaTal: Donein next commit

nagarin (approved):

PR #1345, opened 2021-12-08
ADCM-2377 Speed up dev tests

Sealwing:
This PR is dedicated to speeding up "branch to develop" tests run.

To achieve it two things were done:
1. Some tests were marked as "full" because they take too much time or cover corner cases.
2. API tests were reworked to use "ADCM with dummy data" to skip the phase of pre-filling ADCM that takes about a minute for each test.

Also, the process of choosing which ADCM to use (clean or with dummy data) was reworked so it can be used in API tests the same way as it is used in functional and UI tests.

### DIFF
--- a/tests/functional/test_adcm_upgrade.py
+++ b/tests/functional/test_adcm_upgrade.py
@@ -102,7 +102,7 @@ def _check_encryption(obj: Union[Cluster, Service]) -> None:
 
 
 @pytest.mark.parametrize("adcm_is_upgradable", [True], indirect=True)
-@pytest.mark.parametrize("image", old_adcm_images(), ids=repr, indirect=True)
+@pytest.mark.parametrize("image", old_adcm_images(), ids=repr)
 def test_upgrade_adcm(
     adcm_fs: ADCM,
     sdk_client_fs: ADCMClient,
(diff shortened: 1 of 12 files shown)


dgusakov:
@Sealwing Pls add comprehensive description

tests/functional/test_adcm_upgrade.py
```diff
@@ -121,7 +121,8 @@ def test_upgrade_adcm(
 
 
 @pytest.mark.parametrize("adcm_is_upgradable", [True], indirect=True)
-@pytest.mark.parametrize("image", old_adcm_images(), ids=repr, indirect=True)
+@pytest.mark.parametrize("image", old_adcm_images(), ids=repr)
+@pytest.mark.skip()
```
skhomuti, line 125: remove it

Sealwing: done

tests/functional/test_adcm_upgrade.py
```diff
@@ -102,7 +102,7 @@ def _check_encryption(obj: Union[Cluster, Service]) -> None:
 
 
 @pytest.mark.parametrize("adcm_is_upgradable", [True], indirect=True)
-@pytest.mark.parametrize("image", old_adcm_images(), ids=repr, indirect=True)
+@pytest.mark.parametrize("image", old_adcm_images(), ids=repr)
```
dgusakov, line 105: Why did you remove indirect here?

Sealwing: returned it back

dgusakov (approved):

skhomuti (approved):

PR #1347, opened 2021-12-08
ADCM-2420 Ansible 2.9 as alternative

acmnu:
The commit adds an alternative variant to run playbooks. From now
a bundle developer could set a venv atts on action or on object
to point to a custom venv.

From now we have two venv:

* default - with Ansible 2.8.x
* 2.9 - with Ansible 2.9.x

If venv is ommited in config.yaml, then we have "default" venv.


python/cm/migrations/0081_ansible_venv.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.2.7 on 2021-12-19 17:23
```
a-alferov, line 1: I think this migration file is not needed

acmnu:
bender build

BenderArenadata:

python/cm/migrations/0081_ansible_venv.py
acmnu: Migrations has been merged.

a-alferov: Yes, I will fix it

acmnu:
bender build

BenderArenadata:

a-alferov (approved):

PR #1360, opened 2021-12-13
ADCM-2409: check objects and subjects for policy

nagarin:


python/rbac/services/policy.py
```diff
@@ -64,6 +66,43 @@ def object_policy_add(policy: Policy, obj: ObjectType) -> None:
     policy.object.add(policy_object)
 
 
+def _check_subjects(users: Union[bool, List[User]], groups: Union[bool, List[Group]]) -> None:
```
a-alferov, line 69: I'm not sure if we need this check

a-alferov: not relevant

a-alferov (approved):

PR #1362, opened 2021-12-13
ADCM-2415 Add custom role ADCM settings and upgrade

SofiaTal:


python/api/api_views.py
```diff
@@ -55,16 +55,14 @@ def permission_denied(
     raise exceptions.PermissionDenied(detail=message, code=code)
 
 
-def has_import_perm(user, action_type, model, obj):
-    if user.has_perm(f'cm.{action_type}_clusterbind'):
-        return True
-    if user.has_perm(f'cm.{action_type}_import_of_{model}', obj):
+def has_permission(user, action_type, model, obj):
```
grey-kristy, line 58: I think better name would be   "has_custom_permission"

python/api/api_views.py
```diff
@@ -55,16 +55,14 @@ def permission_denied(
...
+    if user.has_perm(f'cm.{action_type}_of_{model}', obj):
         return True
     return False
 
 
-def check_import_perm(self, action_type, model, obj):
-    if not has_import_perm(self.request.user, action_type, model, obj):
+def check_perm(self, action_type, model, obj):
```
grey-kristy, line 64: here the same - "check_custom_perm"

python/api/provider/views.py
```diff
@@ -96,11 +96,14 @@ def get(self, request, provider_id, upgrade_id):  # pylint: disable=arguments-di
 class DoProviderUpgrade(GenericAPIPermView):
     queryset = Upgrade.objects.all()
     serializer_class = api.serializers.DoUpgradeSerializer
+    check_upgrade_perm = check_perm
```
grey-kristy, line 99: I don't think we should introduce new name here, it doesn't have any additional semantic

grey-kristy (changes requested):

grey-kristy (approved):

PR #1374, opened 2021-12-15
ADCM-2391 fix 500 on /rbac/role without 'child' key

SofiaTal:


python/rbac/services/role.py
```diff
@@ -50,8 +50,11 @@ def role_create(built_in=False, type_of_role=RoleTypes.role, **kwargs) -> Role:
 
 def role_update(role: Role, **kwargs) -> Role:
     """Updating Role object"""
-    child = kwargs.pop('child', None)
-    check_role_child(child)
+    if 'child' in kwargs:
```
nagarin, line 53: it would be more pythonic instead all of that to have single line
child = kwargs.get('child', [])

nagarin (changes requested):

nagarin (approved):

PR #1375, opened 2021-12-15
ADCM-2413 fix flacky error with authorization

grey-kristy:


a-alferov (approved):

a-alferov:
Doesn't make sense because the class has been removed

PR #1382, opened 2021-12-16
ADCM-2434 Delete status_view permission, make status visible

SofiaTal:


python/api/cluster/views.py
```diff
@@ -251,7 +252,7 @@ def post(self, request, cluster_id, upgrade_id):
         return create(serializer, upgrade_id=int(upgrade_id), obj=cluster)
 
 
-class StatusList(GenericAPIPermStatusView, InterfaceView):
+class StatusList(GenericAPIView, InterfaceView):
```
nagarin, line 255: Using GenericAPIView instead of own GenericAPIPermView means access for unauthorized - am I getting that right?

SofiaTal: Using GenericApiView is mean using api_settings.DEFAULT_PERMISSION_CLASSES. In our case it is class IsAuthenticated. Manual testing in API shows that unathorized user getting 401 Unathorized error.

nagarin (approved):

PR #1398, opened 2021-12-20
ADCM-2414 fix role name for component

SofiaTal:


nagarin (approved):

python/cm/bundle.py
```diff
@@ -270,10 +270,17 @@ def cook_roles(bundle):
     for act in Action.objects.filter(prototype__bundle=bundle):
         name = act.display_name
         model = get_model_by_type(act.prototype.type)
-        role_name = (
-            f'{bundle.name}_{bundle.version}_{bundle.edition}_'
-            f'{act.prototype.type}_{act.prototype.display_name}_{name}'
-        )
+        if act.prototype.type == 'component':
```
acmnu, line 273: @SofiaTal I don't like this. Lines 275-277 and 281-282 almost the same. So it is better to write difference to some variable and then compose result from the same template, but with small addition.

acmnu (changes requested):

acmnu (approved):

acmnu:
bender build

BenderArenadata:

PR #1410, opened 2021-12-22
ADCM-2453

kuhella:
fixed duplicates, roles category, naming convention, linked roles base on bundle action to built-in role type roles (Service Administrator etc)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@
 /tests/fixture.db
 /web/node_modules/
 /wwwroot
+/run_adcm.sh
--- a/.gitignore
+++ b/.gitignore
@@ -28,4 +28,3 @@
 /tests/fixture.db
 /web/node_modules/
 /wwwroot
-/run_adcm.sh
--- a/python/cm/bundle.py
+++ b/python/cm/bundle.py
@@ -306,6 +306,7 @@ def cook_roles(bundle):
                     'prototype__bundle_id': bundle.id,
                 },
             },
+            parametrized_by_type=act.prototype.type,
         )
         role.save()
         role.category.add(category)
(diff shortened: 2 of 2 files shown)


python/cm/bundle.py
```diff
@@ -299,32 +306,52 @@ def cook_roles(bundle):
                     'prototype__bundle_id': bundle.id,
                 },
             },
+            parametrized_by_type=act.prototype.type,
```
a-alferov, line 309: `parametrized_by_type` is a `list`

python/cm/bundle.py
```diff
@@ -265,11 +265,18 @@ def update_obj(dest, source, fields):
         setattr(dest, f, getattr(source, f))
 
 
+def extend_role(name, children):
+    role = Role.objects.get(name=name)
+    for child_role in children:
```
a-alferov, line 270: We don't need a loop here, we can add all the children like this `role.child.add(*children)`

a-alferov (changes requested):

python/cm/bundle.py
kuhella: fixed

python/cm/bundle.py
kuhella: fixed

python/cm/bundle.py
nagarin: and yes, hanging comma is bad replacement for explicit list literal

python/cm/bundle.py
```diff
@@ -299,32 +305,52 @@ def cook_roles(bundle):
...
         role.permissions.add(perm)
-
-    for name, children in parent.items():
-        role = Role(
-            name=f'{name}',
-            display_name=f'{name}',
-            description=f'action(s) {name}',
-            bundle=bundle,
+        if name not in parent:
+            parent[name] = {'parametrized_by_type': act.prototype.type, 'children': []}
```
nagarin, line 318: Should'n parametrized_by_type be here a list too?

.gitignore
```diff
@@ -28,3 +28,4 @@
 /tests/fixture.db
 /web/node_modules/
 /wwwroot
+/run_adcm.sh
```
a-alferov, line 31: remove pls

a-alferov (changes requested):

nagarin (approved):

a-alferov (approved):

PR #1418, opened 2021-12-22
ADCM-2054 replace url for the first time

drKaramazin:


web/src/app/abstract-directives/list.directive.ts
```diff
@@ -56,7 +56,9 @@ export abstract class ListDirective extends BaseDirective implements OnInit, OnD
       const list = data.results;
       this.data = new MatTableDataSource<any>(list);
       this.changeCount(data.count);
-      this.listItemEvt.emit({ cmd: 'onLoad', row: list[0] });
+      if (list.length) {
```
remizov-arena, line 59: Is data.results allready array? Why result has any type? May be need `Array.isArray(list)` to condition?

drKaramazin: Yes. It would be more accurate

web/src/app/abstract-directives/list.directive.ts
```diff
@@ -56,7 +56,9 @@ export abstract class ListDirective extends BaseDirective implements OnInit, OnD
       const list = data.results;
       this.data = new MatTableDataSource<any>(list);
       this.changeCount(data.count);
-      this.listItemEvt.emit({ cmd: 'onLoad', row: list[0] });
+      if (Array.isArray(list)) {
+        this.listItemEvt.emit({ cmd: 'onLoad', row: list[0] });
```
remizov-arena, line 60: is allow undefined value? for row

drKaramazin: It seems that such a check is quite enough to fix the error.

remizov-arena (approved):

PR #1426, opened 2021-12-24
ADCM-2507: endpoint for role object candidates

nagarin:


python/rbac/endpoints_ui/role/views.py
```diff
@@ -0,0 +1,105 @@
...
+            models.ObjectType.service.value in role.parametrized_by_type
+            or models.ObjectType.component.value in role.parametrized_by_type
+        ):
+            _services = defaultdict(list)
+            for service in cm_models.ClusterObject.objects.all():
+                _services[service.display_name].append(
+                    {
+                        'name': service.cluster.display_name,
+                        'type': 'service',
+                        'id': service.id,
```
a-alferov, line 87:
```suggestion
                        'type': 'cluster',
                        'id': service.cluster.id,
```

nagarin: No, it is done as required - by selecting cluster from that list user selects instance of that cluster's service

a-alferov: Yes, it dawned on me

kuhella (approved):

PR #1428, opened 2021-12-24
ADCM-2449 add sum of parametrized_by param from child roles

SofiaTal:


python/rbac/services/role.py
```diff
@@ -11,13 +11,35 @@
...
+    role.save()
+
+
+def check_child_parametrized(child_param, param_list) -> bool:
+    cluster_hierarchy = {'cluster', 'service', 'component'}
+    if ((child_param == 'provider') and bool(cluster_hierarchy & set(param_list))) or (
+        (child_param in cluster_hierarchy) and ('provider' in param_list)
+    ):
+        msg = {'This children parametrized by types from different hierarchy'}
+        raise err('ROLE_CREATE_ERROR', msg)
```
a-alferov, line 39: I think the ROLE_ERROR error will be better, since the function is called not only in the create method

SofiaTal: Fixed

a-alferov (approved):

python/rbac/services/role.py
```diff
@@ -11,13 +11,35 @@
...
 from rbac.models import Role, RoleTypes
 from rbac.utils import update_m2m_field
 
 
+def set_parametrized_from_child(role, children: List[Role]):
+    param_list = []
+
+    for child in children:
+        child_params = child.parametrized_by_type
+        for child_param in child_params:
```
nagarin, line 26: It seems a bit overcomplicated. It could be smth like that:
paramset = set()
for ch in children:
    paramset.update(ch.parametrize_by_type)
if paramset.intersection(provider_hierarchy) and paramset.intersection(cluster_hierarchy):
    raise AdwpEx(...)  # not err - look at its definition

nagarin (changes requested):

python/rbac/services/role.py
SofiaTal: Changed

nagarin (approved):

PR #1446, opened 2021-12-29
FIX: use original auth.User.groups for custom user model

nagarin:


acmnu (approved):

nagarin:
user serializer is broken for some reason

a-alferov (approved):

PR #1465, opened 2022-01-11
ADCM-2553 fix PATCH with empty body

SofiaTal:


nagarin (approved):

python/rbac/endpoints/role/views.py
```diff
@@ -32,7 +32,7 @@ class RoleChildSerializer(BaseRelatedSerializer):
 
 class RoleSerializer(FlexFieldsSerializerMixin, serializers.ModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='rbac:role-detail')
-    child = RoleChildSerializer(many=True, required=False)
```
a-alferov, line 35: I think `required=False` should be left

SofiaTal: Fixed

a-alferov (approved):

PR #1472, opened 2022-01-12
ADCM-2444 add tests for predefined roles

skhomuti:


tests/functional/rbac/test_cluster_related_permissions.py
```diff
@@ -411,3 +412,68 @@ def test_remove_bundle(user_policy, user_sdk: ADCMClient, sdk_client_fs):
     delete_policy(user_policy)
     BusinessRoles.UploadBundle.value.method_call(sdk_client_fs)
     is_denied(user_sdk.bundle_list()[-1], BusinessRoles.RemoveBundle)
+
+
+def test_service_administrator(user, user_sdk: ADCMClient, sdk_client_fs, prepare_objects, second_objects):
+    """Test that service administrator role grants access to single service"""
```
Sealwing, line 418: ... and its siblings. maybe clarify it here and below?

Sealwing (approved):

tests/functional/rbac/test_cluster_related_permissions.py
skhomuti: Do you mean components?

Sealwing: Service Admin have access to service and it's components
Cluster Admin to services and their components
Provider to provider and hosts

If a get it right we can generally name all of them siblings, because someday we may ask ourselves: "why it check that access to siblings is allowed, when it's not in docstring?". But maybe it's not a problem idk

PR #1475, opened 2022-01-13
ADCM-2503 Tests on running actions with permissions and rbac tests directory refactoring

Sealwing:


annademina (approved):

tests/functional/rbac/actions/test_run_actions.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def _do_nothing_action_not_presented(obj):
+    """Filter for removing objects that doesn't have "Do nothing" action"""
+    return DO_NOTHING_ACTION not in (a.display_name for a in obj.action_list())
+
+
+# !===== Tests ======!
+
+# pylint: disable-next=too-many-locals
```
skhomuti, line 42: are you sure it's needed here?

tests/functional/rbac/actions/test_run_actions.py
```diff
@@ -0,0 +1,247 @@
...
+        f'Check that granting policy allows to run action "{DO_NOTHING_ACTION}" on {get_object_represent(adcm_object)}'
+    ):
+        (user_object,) = as_user_objects(user_sdk, adcm_object)
+        business_role = action_business_role(adcm_object, DO_NOTHING_ACTION)
+        policy = create_action_policy(admin_sdk, adcm_object, business_role, user=user)
+
+        check_single_action_is_allowed_on_object(DO_NOTHING_ACTION, adcm_object, user_sdk, business_role)
+
+    with allure.step(f"Check that granted permission doesn't allow running '{DO_NOTHING_ACTION}' on other objects"):
+        with user_objects_map(
```
skhomuti, line 87: maybe check actions explicitly without too complex `user_objects_map` func?

skhomuti (changes requested):

tests/functional/rbac/actions/test_run_actions.py
Sealwing: Nope, it's redundant now, thanks

skhomuti: and below too

tests/functional/rbac/actions/test_run_actions.py
Sealwing: At first, I was thinking about `user_objects_map` as a function to be reused to filter different kinds of "denied" objects, but I'm still not sure that it'll be somehow different without this func. I may unwrap it inside the `check_action_is_not_allowed_on_objects`, but do you think it's worth it?

tests/functional/rbac/actions/test_run_actions.py
Sealwing: yep

skhomuti (approved):

PR #1485, opened 2022-01-13
Adcm-2455_5 add ui tests for policies page

annademina:


tests/ui_tests/app/page/admin/locators.py
```diff
@@ -164,3 +164,68 @@ class SelectPermissionsRow:
...
+                By.XPATH, "//div[./mat-option//*[@placeholderlabel='Select role']]/mat-option", "select items for role"
+            )
+
+            users_select = Locator(By.CSS_SELECTOR, "adwp-input-select[label='User'] adwp-select", "select users")
+            users_item = Locator(By.CSS_SELECTOR, "adwp-selection-list mat-list-option", "select items for users")
+
+            group_select = Locator(By.CSS_SELECTOR, "adwp-input-select[label='Group'] adwp-select", "select group")
+            group_item = Locator(By.CSS_SELECTOR, "adwp-selection-list mat-list-option", "select items for group")
+
+            next_btn_first = Locator(
```
Sealwing, line 204: I guess "next_btn" is enough here and below because it's already inside the "{#}Step" container, don't you think so?

Sealwing: Thou description is good, it will be easier to understand the problem in a report

tests/ui_tests/app/page/admin/page.py
```diff
@@ -489,3 +510,83 @@ def click_delete_button(self):
...
+    MENU_SUFFIX = 'policies'
+    MAIN_ELEMENTS = [
+        AdminIntroLocators.intro_title,
+        AdminIntroLocators.intro_text,
+        AdminGroupsLocators.create_group_btn,
+        AdminGroupsLocators.delete_btn,
+        CommonTable.header,
+    ]
+
+    def click_create_policy_btn(self):
```
Sealwing, line 527: I suggest naming it like `open_create_policy_popup` and clear it in docs that it is opened by clicking button

tests/ui_tests/app/page/admin/page.py
```diff
@@ -489,3 +510,83 @@ def click_delete_button(self):
...
+        self.send_text_to_element(AdminPoliciesLocators.AddPolicyPopup.FirstStep.name_input, policy_name)
+        if description:
+            self.send_text_to_element(AdminPoliciesLocators.AddPolicyPopup.FirstStep.description_input, description)
+        with allure.step(f"Select role {role} in popup"):
+            self.find_and_click(AdminPoliciesLocators.AddPolicyPopup.FirstStep.role_select)
+            self.wait_element_visible(AdminPoliciesLocators.AddPolicyPopup.FirstStep.role_item)
+            for available_role in self.find_elements(AdminPoliciesLocators.AddPolicyPopup.FirstStep.role_item):
+                if available_role.text == role:
+                    available_role.click()
+                    break
```
Sealwing, line 545: If the role isn't found, it'll pass silently. I think we need some generic method to search for "text" in a collection of WebElements (maybe error message should be an argument in case entity isn't in collection)

tests/ui_tests/app/page/admin/page.py
```diff
@@ -489,3 +510,83 @@ def click_delete_button(self):
...
+        assert users or groups, "There are should be users or groups in the policy"
```
Sealwing, line 546: I suggest moving this check to the first line of the function and make it not an assert:
```
if not (users or groups):
    raise ValueError("...")
```
The error can be anything else, just an example.

The motivation is following:
1. We validate input at the start and don't try to do things that won't lead us to success anyway.
2. Checking input as the assertion is the right way to do it in the program code, but assertion in our tests will lead to mark test as "having a bug", while not passing users or groups is definitely a test bug

tests/ui_tests/app/page/admin/page.py
```diff
@@ -489,3 +510,83 @@ def click_delete_button(self):
...
+            with allure.step(f"Select groups {groups} in popup"):
+                self.find_and_click(AdminPoliciesLocators.AddPolicyPopup.FirstStep.group_select)
+                self.wait_element_visible(AdminPoliciesLocators.AddPolicyPopup.FirstStep.group_item)
+                available_groups = self.find_elements(AdminPoliciesLocators.AddPolicyPopup.FirstStep.group_item)
+                for group in groups.split(","):
+                    for available_group in available_groups:
+                        if available_group.text == group:
+                            self.scroll_to(available_group)
+                            available_group.click()
+                self.find_and_click(AdminPoliciesLocators.AddPolicyPopup.FirstStep.group_select)
```
Sealwing, line 568: Looks quite the same at the first glance. Maybe we can add a separate function that takes locators for select and item, the text value of user/group, and do stuff?
And I'm not sure what will be if the user/group isn't presented in a list. If it fail silently, it may be difficult to understand and debug

tests/ui_tests/app/page/admin/page.py
```diff
@@ -489,3 +510,83 @@ def click_delete_button(self):
...
+        self.wait_element_hide(AdminPoliciesLocators.AddPolicyPopup.block)
+
+    def get_all_policies(self) -> [AdminGroupInfo]:
+        """Get all policies info."""
+
+        policies_items = []
+        policies_rows = self.table.get_all_rows()
+        for policy in policies_rows:
+            policies_items.append(policy.text)
+        return policies_items
```
Sealwing, line 583: I guess the return type here will be List[str], please clarify your intention

tests/ui_tests/app/page/admin/page.py
```diff
@@ -489,3 +510,83 @@ def click_delete_button(self):
...
+
+    def select_all_groups(self):
+        self.find_elements(self.table.locators.header)[0].click()
```
Sealwing, line 586: Did I get it right: if you click the first table header, it will select all groups? Can you please mention it in the docstring?

tests/ui_tests/app/page/admin/page.py
```diff
@@ -489,3 +510,83 @@ def click_delete_button(self):
...
+
+    def click_delete_button(self):
+        self.find_and_click(AdminPoliciesLocators.delete_btn)
+        self.wait_element_visible(DeleteDialog.body)
+        self.find_and_click(DeleteDialog.yes)
+        self.wait_element_hide(DeleteDialog.body)
```
Sealwing, line 592: As I understand this method will "delete selected policies", please clarify it in the name/docstring

tests/ui_tests/test_admin_page.py
```diff
@@ -347,4 +352,73 @@ def test_delete_group_from_groups_page(self, app_fs):
...
+        users="admin,status",
+        groups=None,
+    )
+
+    def test_open_by_tab_admin_policies_page(self, app_fs):
+        """Test open /admin/policies from left menu"""
+
+        intro_page = AdminIntroPage(app_fs.driver, app_fs.adcm.url).open()
+        policies_page = intro_page.open_policies_menu()
+        policies_page.wait_page_is_opened()
```
Sealwing, line 375: `open_policies_menu` is waiting for the page to be opened, do we need to recheck?

tests/ui_tests/app/page/admin/page.py
Sealwing: And I guess it's selecting all policies, not groups?

tests/ui_tests/test_admin_page.py
```diff
@@ -347,4 +352,73 @@ def test_delete_group_from_groups_page(self, app_fs):
...
+            self.custom_policy.name,
+            self.custom_policy.description,
+            self.custom_policy.role,
+            self.custom_policy.users,
+            self.custom_policy.groups,
+        )
+        policies_page.select_all_groups()
+        policies_page.click_delete_button()
+        with allure.step('Check that policy has been deleted'):
+            assert len(policies_page.table.get_all_rows()) == 1, "There should be 1 default groups"
```
Sealwing, line 424: I've got an impression it is about policies, so I guess it's typos with "select_all_groups" and "There should be 1 default group" :smile_cat:

Sealwing (changes requested):

tests/ui_tests/app/page/admin/page.py
```diff
@@ -62,6 +63,17 @@ class AdminGroupInfo:
     users: str
 
 
+@dataclass
+class AdminPolicyInfo:
+    """Information about poicy"""
```
skhomuti, line 68: `policy` typo

tests/ui_tests/app/page/admin/page.py
```diff
@@ -489,3 +510,83 @@ def click_delete_button(self):
...
+                self.wait_element_visible(AdminPoliciesLocators.AddPolicyPopup.FirstStep.group_item)
+                available_groups = self.find_elements(AdminPoliciesLocators.AddPolicyPopup.FirstStep.group_item)
+                for group in groups.split(","):
+                    for available_group in available_groups:
+                        if available_group.text == group:
+                            self.scroll_to(available_group)
+                            available_group.click()
+                self.find_and_click(AdminPoliciesLocators.AddPolicyPopup.FirstStep.group_select)
+        self.find_and_click(AdminPoliciesLocators.AddPolicyPopup.FirstStep.next_btn_first)
+        self.wait_element_visible(AdminPoliciesLocators.AddPolicyPopup.SecondStep.next_btn_second)
```
skhomuti, line 570: the second step can be different depending on selected role https://arenadata.atlassian.net/wiki/spaces/ADCM/pages/841974322/Requirements+for+ADCM+RBAC.#Add-Policies

skhomuti (changes requested):

tests/ui_tests/app/page/admin/page.py
annademina: changed

tests/ui_tests/app/page/admin/page.py
annademina: moved in the beginning

tests/ui_tests/app/page/admin/page.py
annademina: moved to a func

tests/ui_tests/app/page/admin/page.py
annademina: changed, thnx!

tests/ui_tests/app/page/admin/page.py
annademina: added docstring

tests/ui_tests/test_admin_page.py
annademina: changed, thnx!

tests/ui_tests/app/page/admin/page.py
annademina: yes, I will change it when the time comes)

tests/ui_tests/app/page/admin/page.py
annademina: fixed. thnx!

tests/ui_tests/app/page/admin/page.py
Sealwing: Hm, I still disagree with using assert for what is a test creator error, not a bug. Can you please argue your decision to leave it as an assertion?

tests/ui_tests/app/page/admin/page.py
```diff
@@ -524,47 +524,50 @@ class AdminPoliciesPage(GeneralAdminPage):
...
         with allure.step(f"Select role {role} in popup"):
             self.find_and_click(AdminPoliciesLocators.AddPolicyPopup.FirstStep.role_select)
             self.wait_element_visible(AdminPoliciesLocators.AddPolicyPopup.FirstStep.role_item)
-            for available_role in self.find_elements(AdminPoliciesLocators.AddPolicyPopup.FirstStep.role_item):
+            available_roles = self.find_elements(AdminPoliciesLocators.AddPolicyPopup.FirstStep.role_item)
+            assert role in [a.text for a in available_roles], f"There are no role {role} in select role popup"
+            for available_role in available_roles:
                 if available_role.text == role:
                     available_role.click()
                     break
```
Sealwing, line 548: I think it can be `for ... else` situation? Like "we try to find a role, but if we don't (break wasn't used), we raise assertion"

tests/ui_tests/app/page/admin/page.py
```diff
@@ -524,47 +524,50 @@ class AdminPoliciesPage(GeneralAdminPage):
...
-        assert users or groups, "There are should be users or groups in the policy"
+
+        def fill_users_or_group_select(items, available_items):
+            for item in items.split(","):
+                assert item in [u.text for u in available_items], f"There are no {item} in select role popup"
+                for available_item in available_items:
+                    if available_item.text == item:
+                        self.scroll_to(available_item)
+                        available_item.click()
```
Sealwing, line 556: I think `for ... else` will fit nice here

tests/ui_tests/app/page/admin/page.py
```diff
@@ -573,7 +576,7 @@ def create_policy(
         self.find_and_click(AdminPoliciesLocators.AddPolicyPopup.ThirdStep.create_btn)
         self.wait_element_hide(AdminPoliciesLocators.AddPolicyPopup.block)
 
-    def get_all_policies(self) -> [AdminGroupInfo]:
+    def get_all_policies(self) -> [str]:
```
Sealwing, line 579: Maybe clarify that it's policies' names? Or mention it in docstring

Sealwing (changes requested):

tests/ui_tests/app/page/admin/page.py
annademina: yes. got it) changed

tests/ui_tests/test_admin_page.py
annademina: removed

tests/ui_tests/app/page/admin/page.py
annademina: what situation for else? We check in the 544 line that our item is in the available items and then search this item by text

tests/ui_tests/app/page/admin/page.py
annademina: added

tests/ui_tests/app/page/admin/page.py
Sealwing: I've thought of avoiding iteration over the same collection twice like:
```python
for available_role in available_roles:
    if available_role.text == role:
        available_role.click()
        break
else:
    raise AssertionError(f"There are no role {role} in select role popup")
```
It's a personal taste :cat:, so it's just a suggestion. If you think your way is clearer -- I won't insist.

annademina: changed)

tests/ui_tests/app/page/admin/page.py
annademina: changed

Sealwing (approved):

PR #1503, opened 2022-01-17
ADCM-2417 fix form configuration

drKaramazin:


web/src/app/shared/configuration/services/field.service.ts
```diff
@@ -317,9 +317,9 @@ export class FieldService {
   }
 
   checkValue(value: resultTypes, type: TNForm): resultTypes {
-    if ((value === '' || isEmptyObject(value)) && type === 'map') return {};
+    if ((value === '' || isEmptyObject(value) || value === null) && type === 'map') return {};
```
remizov-arena, line 320: May be will be better ? And not recalc and rewrite conditions
```
if (value === '' || value === null || isEmptyObject(value)) {
  if (type === 'map') return {};
  if (type === 'list') return [];
  return null;
} 
```

remizov-arena (dismissed):

drKaramazin:
bender build

BenderArenadata:
Hey, @drKaramazin! Build has been launched!

a-alferov:
bender build

BenderArenadata:

remizov-arena (approved):

PR #1516, opened 2022-01-19
ADCM-2557 replace two permissions by one object_permission with action n…

SofiaTal:
…ame in codename


python/api/action/views.py
```diff
@@ -151,10 +151,9 @@ def has_action_perm(self, action, obj):
         if action.host_action:
             obj = get_host_object(action, obj.cluster)
 
-        if not user.has_perm('cm.run_object_action', obj):
-            return False
-        if not user.has_perm('cm.run_action', action):
+        if not user.has_perm(f'cm.run_action_{action.name}', obj):
```
nagarin, line 154: return user.has_perm(...)

SofiaTal: Done

nagarin (approved):

PR #1530, opened 2022-01-21
Adcm-2455_6 fix UI admin tests

annademina:


tests/ui_tests/app/page/admin/page.py
```diff
@@ -586,7 +587,17 @@ def get_all_policies(self) -> [str]:
...
+            policy_objects = self.find_child(policy, AdminPoliciesLocators.PolicyRow.objects).text
+            policy_item = AdminPolicyInfo(
+                name=self.find_child(policy, AdminPoliciesLocators.PolicyRow.name).text,
+                description=self.find_child(policy, AdminPoliciesLocators.PolicyRow.description).text,
+                role=self.find_child(policy, AdminPoliciesLocators.PolicyRow.role).text,
+                users=self.find_child(policy, AdminPoliciesLocators.PolicyRow.users).text,
+                groups=policy_groups if policy_groups else None,
+                objects=policy_objects if policy_objects else None,
+            )
+            policies_items.append(policy_item)
```
Sealwing, line 600: we need to fix return type and affected tests

tests/ui_tests/test_admin_page.py
```diff
@@ -382,16 +383,16 @@ def test_create_policy_on_admin_groups_page(self, app_fs):
             self.custom_policy.groups,
         )
         current_policies = policies_page.get_all_policies()
-        with allure.step('Check that there are 2 policies'):
-            assert len(current_policies) == 2, "There should be 2 policies on the page"
-            assert current_policies == ["default", self.custom_policy.name], "Created policy should be on the page"
+        with allure.step('Check that there is 1 policy'):
+            assert len(current_policies) == 1, "There should be 1 policy on the page"
+            assert current_policies == [self.custom_policy], "Created policy should be on the page"
```
Sealwing, line 388: Will they be equal because they're both dataclasses? Because regular objects will have problems with that.
If dataclasses are hashable I suggest use here sets to avoid questions like mine in the future)

Sealwing (changes requested):

tests/ui_tests/test_admin_page.py
Sealwing: works correctly

Sealwing (approved):

PR #1548, opened 2022-01-24
ADCM-2617: re-apply provider policy on host creation

nagarin:


python/rbac/upgrade/role_spec.yaml
```diff
@@ -408,6 +408,8 @@ roles:
       child:
         - Edit host config
         - Edit config endpoints
+        - View host config
```
kuhella, line 411: is it correct? Edit host config already contains view_config_of, why we add it twice?

a-alferov (approved):

PR #1572, opened 2022-01-28
ADCM-2588 UI tests: add categories check to role popup

annademina:


tests/ui_tests/app/page/admin/page.py
```diff
@@ -430,6 +430,11 @@ def fill_description_in_role_popup(self, description: str):
         self.send_text_to_element(AdminRolesLocators.AddRolePopup.description_name_input, description)
         self.find_and_click(AdminRolesLocators.AddRolePopup.role_name_input)
 
+    @allure.step('Fill role categories {categories}')
+    def fill_categories_in_role_popup(self, categories: str):
+        self.send_text_to_element(AdminRolesLocators.AddRolePopup.categories_input, categories)
+        self.find_and_click(AdminRolesLocators.AddRolePopup.categories_input)
```
Sealwing, line 436: Can we add a check here that filtering worked (simply by checking if the amount of roles is decreased)?

Sealwing: or better not not here, but in a separate function

Sealwing (changes requested):

annademina:
won't do

PR #1592, opened 2022-02-02
ADCM-2651 add object view perm, add view perm for business and role roles

SofiaTal:


python/rbac/roles.py
```diff
@@ -151,12 +152,18 @@ def apply(
                     self.find_and_apply(comp, policy, role, user, group)
                 for hc in HostComponent.obj.filter(cluster=obj.cluster, service=obj):
                     self.find_and_apply(hc.host, policy, role, user, group)
+                for cluster in Cluster.obj.filter(id=obj.cluster.id):
+                    self.find_and_apply(cluster, policy, role, user, group)
```
a-alferov, line 156:
```suggestion
                
                self.find_and_apply(obj.cluster, policy, role, user, group)
```
I think the cycle will be superfluous here

SofiaTal: Done

a-alferov (approved):

PR #1597, opened 2022-02-07
ADCM-2292 forbid to upload bundles with import min version higher than max

SofiaTal:


python/cm/stack.py
```diff
@@ -323,6 +324,10 @@ def save_import(proto, conf):
         if 'versions' in conf['import'][key]:
             check_versions(proto, conf['import'][key], f'import "{key}"')
             set_version(si, conf['import'][key])
+            if si.min_version and si.max_version:
+                if rpm.compare_versions(str(si.min_version), str(si.max_version)) > 0:
```
a-alferov, line 328: Why `str()` function ?

SofiaTal: We need to pass str in function. str() helps with version looking like this (1.2.3.4), which can be type ScalarFloat.

alexterent (approved):

a-alferov (approved):

PR #1601, opened 2022-02-08
ADCM-2670 add checks for action and objects existing before its showing in jobs page

SofiaTal:


python/api/job/serializers.py
```diff
@@ -52,23 +52,27 @@ def cook_obj(obj_type, obj_id, name):
...
-                    cluster=cluster, prototype=task.action.prototype.parent
-                )
-                component = ServiceComponent.obj.get(
-                    cluster=cluster, service=service, prototype=task.action.prototype
-                )
-                resp.append(cook_obj('service', service.id, get_object_name(service)))
-                resp.append(cook_obj('component', component.id, get_object_name(component)))
-        else:
-            resp.append(cook_obj('provider', obj.provider.id, get_object_name(obj.provider)))
+        if task.action:
```
nagarin, line 55: May be it would seems more straight-forward before line 49 smth like
if not task.action:
    return resp

python/api/job/serializers.py
```diff
@@ -81,19 +85,20 @@ def get_task_selector(obj, action):
     if obj.prototype.type == 'component':
         selector['cluster'] = obj.cluster.id
         selector['service'] = obj.service.id
-    if isinstance(obj, Host) and action.host_action:
+    if action and isinstance(obj, Host) and action.host_action:
```
nagarin, line 88: and same here

nagarin (approved):

PR #1629, opened 2022-03-02
Adcm-2669 add func tests with reassignment objects in politics

annademina:


tests/functional/rbac/test_reassignment.py
```diff
@@ -149,6 +154,137 @@ def grant_role(self, client: ADCMClient, user: User, role: RbacRoles, *objects:
...
 
+    def test_add_remove_user_from_group_and_policy(self, clients, is_denied_to_user, prepare_objects, user):
+        """Test that policies are applied after user add/remove to the policy"""
+
+        admin_cluster, admin_service, admin_component, *_ = prepare_objects
+
+        is_denied_to_user(admin_cluster, BusinessRoles.EditClusterConfigurations)
+        is_denied_to_user(admin_service, BusinessRoles.EditServiceConfigurations)
+        is_denied_to_user(admin_component, BusinessRoles.EditComponentConfigurations)
+        with allure.step('Create user group'):
```
Sealwing, line 165: Maybe 'create a group with user "test_user"'?

tests/functional/rbac/test_reassignment.py
```diff
@@ -149,6 +154,137 @@ def grant_role(self, client: ADCMClient, user: User, role: RbacRoles, *objects:
...
+            test_group = clients.admin.group_create("Test_group", user=[{"id": user.id}])
+        with allure.step('Create a test policy with user group'):
```
Sealwing, line 167: I suggest 'Create a "Cluster Administrator" policy for a group'. That way we clarify what we grant to which objects (users/groups)

tests/functional/rbac/test_reassignment.py
```diff
@@ -149,6 +154,137 @@ def grant_role(self, client: ADCMClient, user: User, role: RbacRoles, *objects:
...
+                group=[test_group],
+                objects=[admin_cluster],
+            )
+        user_cluster, user_service, user_component = as_user_objects(
+            clients.user, admin_cluster, admin_service, admin_component
+        )
+        is_allowed(user_cluster, BusinessRoles.EditClusterConfigurations)
+        is_allowed(user_service, BusinessRoles.EditServiceConfigurations)
+        is_allowed(user_component, BusinessRoles.EditComponentConfigurations)
+        with allure.step('Change test group: delete user'):
```
Sealwing, line 181: I suggest not using "test" here and in other places. I thinks it's the same as just "group"

tests/functional/rbac/test_reassignment.py
```diff
@@ -149,6 +154,137 @@ def grant_role(self, client: ADCMClient, user: User, role: RbacRoles, *objects:
...
+            test_group.update(user=[])
+        is_denied_to_user(admin_cluster, BusinessRoles.EditClusterConfigurations)
+        is_denied_to_user(admin_service, BusinessRoles.EditServiceConfigurations)
+        is_denied_to_user(admin_component, BusinessRoles.EditComponentConfigurations)
```
Sealwing, line 185: I was thinking about how to wrap and name steps in such cases. What do you think about using the form of "Change _something_ and expect _something_"? e.g. "Delete user from group and expect edit configuration is forbidden on cluster objects"? I think it may be good for grouping purposes

tests/functional/rbac/test_reassignment.py
```diff
@@ -149,6 +154,137 @@ def grant_role(self, client: ADCMClient, user: User, role: RbacRoles, *objects:
...
+        with allure.step('Change test policy: add user'):
+            test_policy.update(user=[{"id": user.id}])
+        is_allowed(user_cluster, BusinessRoles.EditClusterConfigurations)
+        is_allowed(user_service, BusinessRoles.EditServiceConfigurations)
+        is_allowed(user_component, BusinessRoles.EditComponentConfigurations)
+        with allure.step('Change test policy: delete user'):
+            test_policy.update(user=[])
+        is_denied_to_user(admin_cluster, BusinessRoles.EditClusterConfigurations)
+        is_denied_to_user(admin_service, BusinessRoles.EditServiceConfigurations)
+        is_denied_to_user(admin_component, BusinessRoles.EditComponentConfigurations)
```
Sealwing, line 195: Maybe add a case of adding user to the group again?

tests/functional/rbac/test_reassignment.py
```diff
@@ -149,6 +154,137 @@ def grant_role(self, client: ADCMClient, user: User, role: RbacRoles, *objects:
                 name=f'{user.username} is {role.value}', role=client.role(name=role.value), objects=objects, user=[user]
             )
 
+    def test_add_remove_user_from_group_and_policy(self, clients, is_denied_to_user, prepare_objects, user):
+        """Test that policies are applied after user add/remove to the policy"""
```
Sealwing, line 158: A little bit of inconsistency:
the name tells us that the user is added/removed from group and policy
the docstring mentions only policy
in the test body, there's no adding user to the group after policy assignment.

I suggest minor docstring change, like "user add/remove to the policy **directly or as a part of a group**"

tests/functional/rbac/test_reassignment.py
```diff
@@ -149,6 +154,137 @@ def grant_role(self, client: ADCMClient, user: User, role: RbacRoles, *objects:
...
+        is_denied_to_user(first_cluster, role_to_check)
+        is_denied_to_user(second_cluster, role_to_check)
+
+        with allure.step('Create a test policy with user and first cluster'):
+            test_policy = clients.admin.policy_create(
+                name='Test_policy',
+                role=clients.admin.role(name=RbacRoles.ClusterAdministrator.value),
+                user=[user],
+                objects=[first_cluster],
+            )
```
Sealwing, line 218: I guess soon it'll be time to finally make a wrapping for `RbacRoles` based policy creation :D

tests/functional/rbac/test_reassignment_data/provider/actions.yaml
```diff
@@ -0,0 +1,23 @@
...
+- name: "SeeMeInAction"
+  hosts: localhost
+  tasks:
+    - name: "Hehe"
+      assert:
+        that:
+          - 1 == {{job.config.some_param}}
+        msg: "Never gonna give you up"
+      tags:
+        - compare
```
Sealwing, line 23: don't think the action file is required, so I suggest deleting it

Sealwing (changes requested):

tests/functional/rbac/test_reassignment.py
annademina: why?

tests/functional/rbac/test_reassignment.py
annademina: what the difference?

Sealwing: more information on what's in the group

tests/functional/rbac/test_reassignment.py
Sealwing: IMHO "test" doesn't give any more info for the report reader. It does when we have more than one group, naming one "test" and another "berebek" or something :D

tests/functional/rbac/test_reassignment.py
annademina: it is totally not significant what policy do we use in this case, but okay, I'll change, if you think it is

tests/functional/rbac/test_reassignment.py
annademina: IMHO it is a nitpick, but okay, I'll change

tests/functional/rbac/test_reassignment.py
annademina: yes, it is in the 191 line

tests/functional/rbac/test_reassignment_data/provider/actions.yaml
annademina: removed

tests/functional/rbac/test_reassignment.py
annademina: it is a good point, wrapped these cases into allure steps

tests/functional/rbac/test_reassignment.py
annademina: changed

Sealwing (approved):

PR #1671, opened 2022-03-25
ADCM-2728 Rework upgrade function for starting task if upgrade has pending action

SofiaTal:


python/cm/api.py
```diff
@@ -17,6 +17,7 @@
 from django.db import transaction
 from django.utils import timezone
 
+import cm.upgrade
```
a-alferov, line 20: Why?

python/cm/upgrade.py
```diff
@@ -304,7 +305,7 @@ def rpm_cmp_reverse(obj1, obj2):
         return res
 
 
-def do_upgrade(obj: Union[Cluster, HostProvider], upgrade: Upgrade) -> dict:
+def do_upgrade(obj: Union[Cluster, HostProvider], upgrade: Upgrade):
```
a-alferov, line 308:
```suggestion
def do_upgrade(obj: Union[Cluster, HostProvider], upgrade: Upgrade) -> dict:
```

python/cm/upgrade.py
```diff
@@ -313,9 +314,19 @@ def do_upgrade(obj: Union[Cluster, HostProvider], upgrade: Upgrade) -> dict:
...
-    if obj.prototype.type == 'cluster':
+    if not upgrade.action:
+        bundle_switch(obj, upgrade)
+    else:
+        cm.job.start_task(upgrade.action, obj, {}, {}, [], [], False)
+    obj.refresh_from_db()
+    return {'id': obj.id, 'upgradable': bool(get_upgrade(obj))}
+
+
+def bundle_switch(obj: Union[Cluster, HostProvider], upgrade: Upgrade) -> dict:
```
a-alferov, line 325:
```suggestion
def bundle_switch(obj: Union[Cluster, HostProvider], upgrade: Upgrade):
```

a-alferov: no return needed in this function

a-alferov (changes requested):

python/cm/api.py
SofiaTal: I've tryed to fight with cyclic import in linters, but lose :(

a-alferov (approved):

PR #1677, opened 2022-03-29
ADCM-2732 add ui_options and venv to schema

SofiaTal:


a-alferov (changes requested):
I think we should add `ui_options` to the upgrade serializer for UI

PR #1690, opened 2022-04-05
`full` mark added and minor linter-based improvements

Sealwing:


annademina (approved):

tests/ui_tests/test_cluster_list_page.py
```diff
@@ -1238,37 +1247,36 @@ def check_expectations():
...
-                            if expected['alerts'] and not is_read_only:
-                                if field_type == "map":
-                                    is_advanced = cluster_config_page.config.advanced
-                                    cluster_config_page.driver.refresh()
-                                    if is_advanced:
-                                        cluster_config_page.config.click_on_advanced()
-                                    cluster_config_page.config.expand_or_close_group(group_name, expand=True)
-                                else:
-                                    cluster_config_page.config.click_on_advanced()
+                    if not activatable:
```
annademina, line 1250: but in this case we wont check 
assert len(cluster_config_page.config.get_all_config_rows()) == 1, "Field should not be visible"

annademina (changes requested):

tests/ui_tests/test_cluster_list_page.py
Sealwing: Why? This full block was inside of `if not activatable` condition. The line you're refering to won't be checked in previous version if activatable if False either

annademina (approved):

PR #1730, opened 2022-04-28
ADCM-2791 Added remove concerns before delete service

a-alferov:


DanBalalan (approved):

python/cm/api.py
```diff
@@ -344,6 +344,7 @@ def delete_service(service, cancel_tasks=True):
         _cancel_locking_tasks(service)
     service_id = service.id
     cluster = service.cluster
+    service.concerns.filter(type__in=[ConcernType.Issue, ConcernType.Flag]).delete()
```
SofiaTal, line 347: Why not cm.issue.update_hierarchy_issues(service)?

PR #1732, opened 2022-04-28
ADCM-2799 set disabling cause 'mm' to actions having parents's host(s) in mm

DanBalalan:


python/api/utils.py
```diff
@@ -119,6 +119,13 @@ def set_disabling_cause(obj: ADCMEntity, action: Action) -> None:
         ).exists()
         if not action.allow_in_maintenance_mode and mm:
             action.disabling_cause = 'maintenance_mode'
+    elif obj.prototype.type == 'host':
+        mm = HostComponent.objects.filter(
+            component_id__in=HostComponent.objects.filter(host=obj).values_list('component_id'),
+            host__maintenance_mode=MaintenanceModeType.On,
```
a-alferov, line 125: Add filter `host__host_action=False`

a-alferov (approved):

PR #1745, opened 2022-05-10
ADCM-2781 Fixed aggregation statuses for cluster and service with hosts in maintenance mode

a-alferov:


a-alferov:
bender build

BenderArenadata:

DanBalalan (approved):

go/adcm/status/handlers.go
```diff
@@ -295,10 +295,67 @@ func showHost(h Hub, w http.ResponseWriter, r *http.Request) {
...
-	val, _ := h.HostStorage.get(ALL, hostId)
+	val, _ := h.HostStatusStorage.get(ALL, hostId)
 	jsonOut(w, r, val)
 }
 
+func listHost(h Hub, w http.ResponseWriter, r *http.Request) {
+	allow(w, "GET, POST")
+	result, _ := h.ListHostStorage.list()
+	jsonOut(w, r, result)
+
```
Sealwing, line 306: Is there a reason to have empty lines here?

go/adcm/status/handlers.go
```diff
@@ -295,10 +295,67 @@ func showHost(h Hub, w http.ResponseWriter, r *http.Request) {
...
+}
+
+func createHost(h Hub, w http.ResponseWriter, r *http.Request) {
```
Sealwing, line 309: Name "createHost" can mean a lot of things, please be more specific and maybe add a docstring

go/adcm/status/handlers.go
```diff
@@ -295,10 +295,67 @@ func showHost(h Hub, w http.ResponseWriter, r *http.Request) {
...
+	_, ok = h.ServiceMap.getHostCluster(hostId)
+	if !ok {
+		ErrOut4(w, r, "HOST_NOT_FOUND", "unknown host")
+		return
+	}
+	value, _ := h.HostStorage.retrieve(hostId)
+	jsonOut(w, r, value)
+}
+
+func updateHost(h Hub, w http.ResponseWriter, r *http.Request) {
```
Sealwing, line 338: Maybe "updateHostInfo"? Because in statusserver we may update host status, component on host status, host's MM and maybe even host's presence in a cluster

go/adcm/status/handlers.go
```diff
@@ -295,10 +295,67 @@ func showHost(h Hub, w http.ResponseWriter, r *http.Request) {
...
+	allow(w, "GET, PUT")
+	hostId, ok := getPathId(w, r, "hostid")
+	if !ok {
+		return
+	}
+	host := Host{}
+	_, err := decodeBody(w, r, &host)
+	if err != nil {
+		return
```
Sealwing, line 347: Return error with 400 code?

go/adcm/status/host.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (hs *HostStorage) run() {
+	logg.I.f("start storage %s server", hs.label)
+	for {
+		request := <-hs.in
+		logg.I.f("Storage %s command: %+v", hs.label, request.command)
+		switch request.command {
+		case "retrieve":
+			value, ok := hs.db.retrieve(request.id)
+			hs.out <- hostResponse{host: value, ok: ok}
```
Sealwing, line 46: Maybe call an update if not `ok`? Or creating new object each time we have no info on a host is okay in this case?
When don't we know about the host anyway?

go/adcm/status/host.go
```diff
@@ -0,0 +1,169 @@
...
+	return Host{Id: id, MaintenanceMode: false}, false
+}
+
+func (db dbHost) update(id int, maintenance_mode bool) int {
+	code := 200
+	host, ok := db[id]
+	if !ok {
+		db[id] = Host{Id: id, MaintenanceMode: maintenance_mode}
+		code = 201
+	}
```
Sealwing, line 85: it looks like `update_or_create`, but maybe something like `setHostInfo` with a little docstring?

go/adcm/status/host.go
```diff
@@ -0,0 +1,169 @@
...
+	return code
+}
+
+func (db dbHost) create(hosts []Host) int {
+	db.clear()
+	for _, host := range hosts {
+		db[host.Id] = host
+	}
+	return 201
+}
```
Sealwing, line 98: Isn't it recreating `dbHost`, filling it with new instances? From signature one couldn't tell if it's replacing or just "adding" despite the name

go/adcm/status/host.go
```diff
@@ -0,0 +1,169 @@
...
+	ok    bool
+	hosts []Host
+}
+
+type ListHostStorage struct {
+	in    chan listHostRequest
+	out   chan listHostResponse
+	db    dbHost
+	label string
+}
```
Sealwing, line 122: I think you may add another channels to the `HostStorage`:
`single-in` / `single-out`
`multi-in` / `multi-out`
or something like that.

go/adcm/status/host.go
```diff
@@ -0,0 +1,169 @@
...
+	return response.code
+}
+
+func (db dbHost) list() ([]Host, bool) {
+	result := make([]Host, 0)
+	for _, host := range db {
+		result = append(result, host)
+	}
+	return result, true
+}
```
Sealwing, line 169: I think it should be close to another "methods" of `dbHost`

Sealwing (changes requested):

Sealwing:
As I understand, there are updates of host info on host add/remove to/from cluster as well as map-unmap, etc?

go/adcm/status/host.go
Sealwing: With `in` it won't work, but we may just extend request object or make it nested

go/adcm/status/handlers.go
a-alferov: fixed

go/adcm/status/host.go
a-alferov: fixed

go/adcm/status/host.go
a-alferov: fixed

go/adcm/status/host.go
a-alferov: fixed

go/adcm/status/handlers.go
Sealwing: I suggest mention here deletion of previously existed hosts.
And typo in word "Format"

go/adcm/status/host.go
```diff
@@ -76,6 +108,7 @@ func (db dbHost) retrieve(id int) (Host, bool) {
 	return Host{Id: id, MaintenanceMode: false}, false
 }
 
+// update - updated or created Host entity, implements the PUT method
```
Sealwing, line 111: Noticed one thing now:
this is a method of "low-level" structure and it returns "code" from API level instead of flag "updated/created"

I suggest trying to avoid such a mixup in further work, so we can clearly see what and where is set (fact of creation set where it's updated/created, but response code is detected on API level).

I suggest "update or create" instead of past tence, because it's what this method will do when you call it, not what it has already done when it was defined

go/adcm/status/host.go
```diff
@@ -89,6 +122,7 @@ func (db dbHost) update(id int, maintenance_mode bool) int {
 	return code
 }
 
+// create - clear db and created Host entities
```
Sealwing, line 125: same with past tence

Sealwing (changes requested):

go/adcm/status/handlers.go
a-alferov: fixed

go/adcm/status/host.go
a-alferov: fixed

go/adcm/status/host.go
a-alferov: fixed

go/adcm/status/handlers.go
```diff
@@ -300,14 +300,15 @@ func showHost(h Hub, w http.ResponseWriter, r *http.Request) {
...
 func listHost(h Hub, w http.ResponseWriter, r *http.Request) {
 	allow(w, "GET, POST")
-	result, _ := h.HostStorage.list()
+	result := h.HostStorage.list()
 	jsonOut(w, r, result)
 }
 
 // createHost - POST method for create Host entities
-// Foramt example: [{"id": 1, "maintenance_mode": false}, ...]
+// Request foramt: [{"id": 1, "maintenance_mode": false}, ...]
```
Sealwing, line 311: "format", not "foramt" :D

go/adcm/status/handlers.go
Sealwing: hm, I guess I've seen it's called directly even with `decodeBody`..., ok

go/adcm/status/host.go
```diff
@@ -105,30 +104,27 @@ func (db dbHost) retrieve(id int) (Host, bool) {
...
 		return value, true
 	}
-	return Host{Id: id, MaintenanceMode: false}, false
+	return Host{}, false
 }
 
-// update - updated or created Host entity, implements the PUT method
-func (db dbHost) update(id int, maintenance_mode bool) int {
-	code := 200
+// update - update or create Host entity
```
Sealwing, line 110: You don't create entity now

a-alferov: fixed

go/adcm/status/handlers.go
a-alferov: fixed

Sealwing (approved):

PR #1758, opened 2022-05-13
Adcm-2751 add ui test for "Manage Maintenance mode" role

annademina:


tests/ui_tests/app/page/host_list/page.py
```diff
@@ -178,6 +178,38 @@ def _check_host_state(page: HostListPage, row: WebElement):
...
+            if allow_mm_state:
+                assert "mat-primary" in button_state, "Button should be gray"
+                assert (
+                    "Turn maintenance mode ON" in tooltips_info
+                ), "There should be tooltip that user could turn on maintenance mode"
+            else:
+                assert "mat-on" in button_state, "Button should be red"
+                assert (
+                    "Turn maintenance mode OFF" in tooltips_info
+                ), "There should be tooltip that user could turn off maintenance mode"
```
Sealwing, line 208: What does `allow_mm_state` mean?

tests/ui_tests/test_admin_page.py
```diff
@@ -822,3 +824,30 @@ def test_policy_permission_to_run_cluster_action_and_view_task(
...
+            )
+        with allure.step("Create test policy"):
+            sdk_client_fs.policy_create(
+                name="Test policy",
+                role=test_role,
+                user=[sdk_client_fs.user(username=another_user['username'])],
+                objects=[host],
+            )
+        login_page = LoginPage(app_fs.driver, app_fs.adcm.url).open()
+        login_page.login_user(**another_user)
```
Sealwing, line 847: Why do you want to explicitly login via UI?

Sealwing (changes requested):

tests/ui_tests/app/page/host_list/page.py
annademina: renamed as is_mm_state_on

tests/ui_tests/test_admin_page.py
annademina: yep, changed to login by api

tests/ui_tests/app/page/host_list/page.py
Sealwing: Then what's about the "disabled" state? If it shouldn't be checked by this method, then please note that in docstring.
Otherwise you may want to use constants from func tests' conftest file (maybe move it to an upper level), so we have similar API.

tests/ui_tests/test_admin_page.py
Sealwing: Good, but why do you want to login not in fixture?
Is there a reason to do it after the policies are created?

annademina: need to relogin by new user after new policy creation

Sealwing: or reenter the page?
Do you mean that permissions are read only on login?
hmm...

tests/ui_tests/app/page/host_list/page.py
annademina: added for disable state

tests/ui_tests/test_admin_page.py
annademina: i need to change user and check that policy is working for this user

Sealwing: got it, okay

Sealwing (approved):

PR #1767, opened 2022-05-18
ADCM-2752 add ui tests for maintenance mode

annademina:


tests/ui_tests/app/page/cluster/page.py
```diff
@@ -445,6 +448,50 @@ def delete_host_by_row(self, row: WebElement):
...
+        """
+        Assert maintenance mode state in row
+        :param row_num: number of the row with maintenance mode button
+        :param is_mm_state_on: state of maintenance mode button:
+            True for ON state
+            False for OFF state
+            None for not available state
+        """
+
+        def _check_mm_state(page: ClusterHostPage, row: WebElement):
```
Sealwing, line 469: since you define this function inside another, you can omit passing page and row, but get them from the method scope directly

Sealwing (approved):

PR #1788, opened 2022-05-30
ADCM-2780 - Rework host disabling logic on HC page and on HC of action dialog menu

dbardin90:
https://arenadata.atlassian.net/browse/ADCM-2780

### DIFF
--- a/web/src/app/shared/host-components-map/much-2-many/much-2-many.component.ts
+++ b/web/src/app/shared/host-components-map/much-2-many/much-2-many.component.ts
@@ -11,7 +11,7 @@
 // limitations under the License.
 import { AfterViewChecked, ChangeDetectorRef, Component, EventEmitter, Input, Output } from '@angular/core';
 import { FormGroup } from '@angular/forms';
-import { isNumber } from '@app/core/types';
+import { IActionParameter, isNumber } from '@app/core/types';
 import { CompTile, Tile } from '../types';
 import { StatusType } from "@app/components/maintenance-mode-button/maintenance-mode-button.component";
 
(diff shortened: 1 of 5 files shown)


web/src/app/shared/host-components-map/much-2-many/much-2-many.component.ts
```diff
@@ -38,6 +41,14 @@ export class Much2ManyComponent implements AfterViewChecked {
     return this.model.disabled;
   }
 
+  isHostDisabled() {
+    if (this.model?.mm === this.statusType.On) {
```
remizov-arena, line 45: hmm, may be? )))
```
return this.model?.mm === this.statusType.On
```

web/src/app/shared/host-components-map/much-2-many/much-2-many.component.ts
```diff
@@ -82,4 +93,12 @@ export class Much2ManyComponent implements AfterViewChecked {
       lim = isNumber(b) ? b : a === 'odd' ? 1 : a === 'depend' ? 0 : a;
     return `${this.model.relations.length}${lim !== 0 ? ` / ${lim}` : ''}`;
   }
+
+  tooltip() {
+    if (this.model?.mm === this.statusType.On) {
```
remizov-arena, line 98: I think, we can use `this.isHostDisabled()`

web/src/app/shared/host-components-map/much-2-many/much-2-many.component.ts
dbardin90: Thanks, I forgot that condition was simplified

remizov-arena (approved):

PR #1799, opened 2022-06-06
ADCM-2862 exclude concerns from host in mm from component(etc) hierarchy

SofiaTal:


DanBalalan (dismissed):

python/cm/hierarchy.py
```diff
@@ -163,6 +164,7 @@ def _build_tree_up(self, node: Node) -> None:
             parent_values = [
                 hc.component
                 for hc in HostComponent.objects.filter(host=node.value)
+                .exclude(host__maintenance_mode=MaintenanceModeType.On.value)
```
a-alferov, line 167: `.value` not necessary

SofiaTal: Fixed

DanBalalan (approved):

a-alferov (approved):

a-alferov (approved):

a-alferov (commented):
We are waiting for tests from #1802 and #1804

PR #1813, opened 2022-06-15
Adcm 2890

DanBalalan:


SofiaTal (approved):

a-alferov (approved):

python/rbac/ldap.py
```diff
@@ -0,0 +1,116 @@
+import os
```
Sealwing, line 1: don't forget to add license text

DanBalalan: done

Sealwing (approved):

PR #1830, opened 2022-06-23
ADCM-2899 add python scripts to synchronize ldap user and groups and test connection

SofiaTal:


conf/adcm/python_scripts/run_ldap_sync.py
```diff
@@ -0,0 +1,142 @@
+import os
```
Sealwing, line 1: don't forget license :D

DanBalalan (approved):

PR #1835, opened 2022-06-28
ADCM-2922 fix default value to logrotation cnfiguration maxsize

SofiaTal:


conf/adcm/config.yaml
```diff
@@ -79,9 +79,8 @@
           display_name: "Max file size"
           description: |
             Specifies the allowed size (in Mb) the log file can reach before it is archived
```
DanBalalan, line 81: I think we need to remove "(in Mb)" from the description

DanBalalan:
to properly generate logrotate.conf file we need to remove "M" from `size` field in `conf_file_args` in `python/cm/management/commands/logrotate.py`

conf/adcm/config.yaml
SofiaTal: Done, also fix arguments in conf_file_args

DanBalalan (approved):

DanBalalan (approved):

a-alferov (approved):

PR #1840, opened 2022-07-04
ADCM-2941 recheck issues in get_upgrade

DanBalalan:


python/cm/upgrade.py
```diff
@@ -272,6 +272,7 @@ def rpm_cmp_reverse(obj1, obj2):
         upg.license = upg.bundle.license
         res.append(upg)
 
+    cm.issue.update_hierarchy_issues(obj)
```
a-alferov, line 275: I think this code should be added to line 307

a-alferov (approved):

PR #1861, opened 2022-07-12
ADCM-2952 add deactivate users during sync, change error message in case of wrong connection

SofiaTal:


DanBalalan (approved):

conf/adcm/python_scripts/run_ldap_sync.py
```diff
@@ -42,7 +42,7 @@ def _bind(self):
         try:
             l.simple_bind_s(self.settings["BIND_DN"], self.settings["BIND_PASSWORD"])
         except ldap.LDAPError as e:
-            print("Error connecting to %s: %s" % (self.settings["BIND_DN"], e))
+            print(f"Can't connect to {self.settings['SERVER_URI']} with user: {self.settings['BIND_DN']}. Error: {e}")
```
a-alferov, line 45: I think it's better to use here `sys.stderr`

conf/adcm/python_scripts/run_ldap_sync.py
```diff
@@ -159,8 +162,9 @@ def _sync_ldap_users(self, ldap_users):
         django_usernames = set(User.objects.filter(type=OriginType.LDAP).values_list('username', flat=True))
         for username in django_usernames - ldap_usernames:
             user = User.objects.get(username__iexact=username)
-            print(f"We will delete this user: {user}")
-            user.delete()
+            print(f"We will delete this user and deactivate its session: {user}")
```
a-alferov, line 165: I think it's better to use here `sys.stdout`

a-alferov (changes requested):

PR #1868, opened 2022-07-12
ADCM-2964 auth only users in groups specified by group_search

DanBalalan:


python/rbac/ldap.py
```diff
@@ -43,8 +42,38 @@ def _process_extra_filter(filterstr: str) -> str:
         return filterstr
 
 
-def _get_ldap_default_settings():
+# pylint: disable=inconsistent-return-statements
```
winmasta, line 45: It's better to make return statements consistent

python/rbac/ldap.py
```diff
@@ -98,40 +127,36 @@ def _get_ldap_default_settings():
...
 
 class CustomLDAPBackend(LDAPBackend):
+    default_settings = {}
+    is_tls = False
+
     def authenticate_ldap_user(self, ldap_user, password):
         self.default_settings = _get_ldap_default_settings()
         if not self.default_settings:
             return None
+        self.is_tls = is_tls(self.default_settings['SERVER_URI'])
```
winmasta, line 151: self property needs to be set in __init__ as well as default_settings and is_tls

python/rbac/ldap.py
```diff
@@ -143,25 +168,54 @@ def authenticate_ldap_user(self, ldap_user, password):
...
+        username = ldap_user._username  # pylint: disable=protected-access
+
         if User.objects.filter(username__iexact=username, type=OriginType.Local).exists():
             log.exception('usernames collision: `%s`', username)
-            raise AdcmEx('LDAP_USERNAMES_COLLISION')
+            return False
+
+        group_member_attr = self.default_settings['GROUP_TYPE'].member_attr
+        for _, group_attrs in self.__get_groups_by_group_search():
+            if user_dn.lower() in [i.lower() for i in group_attrs.get(group_member_attr, [])]:
```
winmasta, line 210: group_attrs.get(group_member_attr, []) will return empty list if there is no group_member_attr key in group_attrs but list don't have `lower` method

python/rbac/ldap.py
```diff
@@ -143,25 +168,54 @@ def authenticate_ldap_user(self, ldap_user, password):
...
+                break
+        else:
+            return False
+
+        return True
 
     @staticmethod
-    def __det_ldap_group_dn(group_name: str, ldap_groups: list) -> str:
+    def __get_ldap_group_dn(group_name: str, ldap_groups: list) -> str:
         return [i for i in ldap_groups if i[0] == group_name][0][1]
```
winmasta, line 219: IndexError may occurs here

python/rbac/ldap.py
DanBalalan: fixed in https://github.com/arenadata/adcm/pull/1874

python/rbac/ldap.py
DanBalalan: fixed in https://github.com/arenadata/adcm/pull/1874

python/rbac/ldap.py
DanBalalan: fixed in https://github.com/arenadata/adcm/pull/1874

PR #1880, opened 2022-07-18
ADCM-2990 move applying new hc after upgrade in case of hc problem

SofiaTal:


SofiaTal:
bender build

BenderArenadata:
Hey, @SofiaTal! Build has been launched!

DanBalalan (approved):

PR #1887, opened 2022-07-19
ADCM-2972 Add logic for write audit operations

winmasta:


python/audit/utils.py
```diff
@@ -0,0 +1,59 @@
...
+        view: View = args[0]
+        audit_operation = AUDIT_OPERATION_MAP.get(view.__class__.__name__)
+        object_changes = {}
+
+        if is_success(status_code):
+            operation_result = AuditLogOperationResult.Success
+            if resp.data:
+                audit_object = AuditObject.objects.create(
+                    object_id=resp.data.serializer.instance.id,
+                    object_name=resp.data.serializer.instance.name,
```
Sealwing, line 36: I think there won't be `name` for host entities, it'll be `fqdn`

winmasta: It'll be fixed after hosts audit implementation

Sealwing: okay
we practice resolving comments by author

python/audit/tests/test_bundle.py
```diff
@@ -0,0 +1,80 @@
...
+
+        self.audit_operation_upload_bundle = AUDIT_OPERATION_MAP["UploadBundle"]["POST"]
+        self.audit_operation_load_bundle = AUDIT_OPERATION_MAP["LoadBundle"]["POST"]
+
+    def test_upload_success(self):
+        self.upload_bundle()
+
+        log: AuditLog = AuditLog.objects.first()
+
+        assert not log.audit_object
```
kuhella, line 27: is this mean that we expect object is absent? in general it is debatable cause we set object_type, but not set an object

python/audit/models.py
```diff
@@ -68,3 +70,52 @@ class AuditSession(models.Model):
...
+            f"{AuditLogOperationType.Create.label}d",
+            operation_type=AuditLogOperationType.Create.label,
+            object_type=AuditObjectType.Cluster.label,
+        ),
+    },
+    "ConfigLogViewSet": {
+        "POST": AuditOperation(
+            name="???",
+            operation_type=AuditLogOperationType.Create.label,
+            object_type="???",
```
kuhella, line 116: it is depend on prototype type

python/audit/models.py
```diff
@@ -68,3 +70,52 @@ class AuditSession(models.Model):
...
+        "POST": AuditOperation(
+            name=f"{AuditObjectType.Cluster.label.capitalize()} "
+            f"{AuditLogOperationType.Create.label}d",
+            operation_type=AuditLogOperationType.Create.label,
+            object_type=AuditObjectType.Cluster.label,
+        ),
+    },
+    "ConfigLogViewSet": {
+        "POST": AuditOperation(
+            name="???",
```
kuhella, line 114: which endpoint is this?

python/audit/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    def wrapped(*args, **kwargs):
+        error = None
+
+        try:
+            resp = func(*args, **kwargs)
+            status_code = resp.status_code
+        except AdcmEx as exc:
+            error = exc
+            resp = None
+            status_code = exc.status_code
```
kuhella, line 25: might be add additional logging for this situation?

kuhella (commented):
1. what you think about usage parameters in decorator instead of AUDIT_OPERATION_MAP?
2. May be add some tests cases with deleting object to be sure that is_deleted sets True?

python/api/cluster/serializers.py
```diff
@@ -299,25 +321,36 @@ class BindSerializer(serializers.Serializer):
     import_service_id = serializers.SerializerMethodField()
     import_service_name = serializers.SerializerMethodField()
 
-    def get_export_cluster_prototype_name(self, obj):
+    def update(self, instance, validated_data):
+        pass
+
+    def create(self, validated_data):
+        pass
```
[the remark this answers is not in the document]
winmasta: It fixes warning "Class must implement all abstract methods"

python/audit/tests/test_bundle.py
winmasta: If we have no object we don't create record in audit object table. Or we have to create audit object in any case ?

python/audit/utils.py
winmasta: This situation can occurs when something went wrong in View code, so i believe it must be some logging there, not here. Here we should log things only about audit itself

SofiaTal (commented):
Кажется нужно добавить еще лицензию к новым файлам. Не обязательно здесь, можно в новом pr

SofiaTal (approved):

PR #1896, opened 2022-07-21
ADCM-2848_1 add tests to test_run_upgrade_v2_on_cluster_list_page

annademina:


tests/ui_tests/app/page/cluster_list/page.py
```diff
@@ -34,6 +37,9 @@
     ActionDialog,
     DeleteDialog,
 )
+from tests.ui_tests.app.page.common.popups.locator import (
+    HostCreationLocators,
+)
```
Sealwing, line 42: Is it some instrument's work of doing this? :D

tests/ui_tests/app/page/cluster_list/page.py
```diff
@@ -172,6 +178,7 @@ def run_upgrade_in_cluster_row(
         upgrade_name: str,
         config: Optional[dict] = None,
         hc_acl: bool = False,
+        is_new_host: bool = False,
```
Sealwing, line 181: please describe this parameter's purpose in docstring:
1. What it will do (create new host in complex upgrade popup, I guess)
2. Interaction with other params (it is dependant on `hc_acl` and we need to mention that there's no point of passing `is_new_host` without `hc_acl=True`

tests/ui_tests/app/page/cluster_list/page.py
```diff
@@ -193,6 +200,11 @@ def run_upgrade_in_cluster_row(
             if self.is_element_displayed(ActionDialog.next_btn):
                 self.find_and_click(ActionDialog.next_btn)
         if hc_acl:
+            if is_new_host:
+                cluster_components_page = ClusterComponentsPage(self.driver, self.base_url, 1)
+                cluster_components_page.click_add_host_btn()
+                cluster_components_page.host_popup.create_host("Test_host")
+                self.wait_element_hide(HostCreationLocators.block)
```
Sealwing, line 207: It's very confusing why we use another page to run something.

Maybe we may move `click_add_host_btn` to some generic class/function that can take driver and maybe some info on which locator to use (like `in_context: Literal["popup", "cluster-components", "host-list", ...]`) and inside the function there'll be routing based on case.

Maybe it should be like `HostCreator` or something like that, so it'll have public methods for each case (e.g. `on_host_list_page`, `on_upgrade_popup`, `on_cluster_components_page`) and fewer private implementation functions.

What do you think?

tests/ui_tests/test_cluster_list_page.py
```diff
@@ -394,19 +394,32 @@ def test_run_upgrade_on_cluster_list_page(self, sdk_client_fs, app_fs):
...
-            ("upgrade_with_hc_acl_and_disclaimer", None, True, DISCLAIMER_TEXT),
-            ("upgrade_with_config_and_disclaimer", {"somestring2": "test"}, False, DISCLAIMER_TEXT),
-            ("upgrade_with_config_and_hc_acl_and_disclaimer", {"somestring2": "test"}, True, DISCLAIMER_TEXT),
+            ("simple_upgrade", None, False, False, False),
+            ("upgrade_with_hc_acl", None, True, False, False),
+            ("upgrade_with_config", {"somestring2": "test"}, False, False, False),
+            ("upgrade_with_config_and_hc_acl", {"somestring2": "test"}, True, False, False),
+            ("upgrade_with_hc_acl_and_disclaimer", None, True, False, DISCLAIMER_TEXT),
+            ("upgrade_with_config_and_disclaimer", {"somestring2": "test"}, False, False, DISCLAIMER_TEXT),
+            ("upgrade_with_config_and_hc_acl_and_disclaimer", {"somestring2": "test"}, True, False, DISCLAIMER_TEXT),
```
Sealwing, line 405: maybe make it like 1 arg?
dict with key-values that are really required to be passed to `run_upgrade_in_cluster_row` as `run_upgrade_in_cluster_row(**upgrade_kwargs)`, then it will be easier to read (no need to connection arg position and name in the string above)

Sealwing (changes requested):

tests/ui_tests/app/page/cluster_list/page.py
annademina: no, it's just pycharm) in case i need to import smthg else

tests/ui_tests/test_cluster_list_page.py
annademina: not shure, for me it looks like this version is more readable

tests/ui_tests/app/page/cluster_list/page.py
annademina: added

tests/ui_tests/app/page/cluster_list/page.py
annademina: yep, that's good idea. moved to special class everything about host-component

tests/ui_tests/app/page/cluster_list/page.py
Sealwing: :D

tests/ui_tests/test_cluster_list_page.py
Sealwing: okay, tomatos tomahtos)

Sealwing (approved):

PR #1921, opened 2022-07-28
ADCM-3008 add check of hc_acl constraints before pre-upgrade step with prototype from new bundle

SofiaTal:


kuhella (approved):
I test t it work fine!

PR #1930, opened 2022-07-29
Fix extra concern in case of [0,1] constraint

SofiaTal:


SofiaTal:
bender build

BenderArenadata:
Hey, @SofiaTal! Build has been launched!

PR #1942, opened 2022-08-04
ADCM-2972 Add update audit

winmasta:


winmasta:
Bender build

BenderArenadata:
Hey, @winmasta! Build has been launched!

python/audit/utils.py
```diff
@@ -86,6 +87,15 @@ def _task_case(task_pk: str, action: str) -> Tuple[AuditOperation, AuditObject]:
     return audit_operation, audit_object
 
 
+def _get_service_name(service: ClusterObject) -> str:
+    if service.display_name:
+        name = service.display_name
+    else:
+        name = str(service)
+
+    return name
```
a-alferov, line 96: I think this `else` is redundant. And need add `service` to `display_name` or use `service.prototype.name` instead `str(service)`:
```suggestion
def _get_service_name(service: ClusterObject) -> str:
    if service.display_name:
        return service.display_name
    return service.prototype.name
```

a-alferov (changes requested):

python/audit/utils.py
winmasta: `service.prototype.name` is nullable so fixed with addition

a-alferov (approved):

PR #1953, opened 2022-08-09
ADCM-3047 Fix service and component audit log object type

winmasta:


python/api/audit/urls.py
```diff
@@ -11,15 +11,10 @@
...
 
-from .views import AuditOperationViewSet, AuditLoginViewSet, AuditRoot
-
 router = SimpleRouter()
-router.register(r'operation', AuditOperationViewSet, basename='audit-operations')
-router.register(r'login', AuditLoginViewSet, basename='audit-logins')
-urlpatterns = [
-    path(r'', AuditRoot.as_view(), name='root'),
-    *router.urls,
-]
```
Sealwing, line 25: As I understand, without registering `AuditRoot`, it'll not be shown in viewer in API root, but operation/ and login/ will be available through direct access.
@a-alferov Do we want that?
It looks like it doesn't break ADCM Client thou

winmasta: Reverted

Sealwing (approved):

PR #1962, opened 2022-08-11
ADCM-3046 Add authentication audit for `login/` endpoint

a-alferov:


python/audit/middleware.py
```diff
@@ -10,30 +10,22 @@
...
-            return self.get_response(request)
-
-        body = json.loads(request.body.decode('utf-8'))
-        username = body['username']
-        response = self.get_response(request)
-
-        if request.user.is_authenticated:
-            audit_user = request.user
+    @staticmethod
+    def _audit(user: User | AnonymousUser = None, username: str = None):
```
winmasta, line 23: Maybe `Optional` needs to be used if `None` is default

python/audit/middleware.py
```diff
@@ -10,30 +10,22 @@
...
+        """Authentication audit"""
+        if user.is_authenticated:
```
winmasta, line 25: If `user` is `None` `AttributeError` will be raised

python/audit/middleware.py
```diff
@@ -43,7 +35,18 @@ def __call__(self, request):
...
                 result = AuditSessionLoginResult.UserNotFound
+                user = None
+
+        AuditSession.objects.create(user=user, login_result=result, login_details=details)
+
+    def __call__(self, request):
+        response = self.get_response(request)
 
-        AuditSession.objects.create(user=audit_user, login_result=result, login_details=details)
+        if request.method == "POST" and request.path in [
```
winmasta, line 45: nit: find in set is faster than in list

python/audit/middleware.py
a-alferov: fixed

a-alferov: I use `Optional`, but maybe use `|`? For example `User | AnonymousUser | None`

python/audit/middleware.py
a-alferov: fixed

python/audit/middleware.py
a-alferov: fixed

winmasta (approved):

a-alferov:
bender build

BenderArenadata:

winmasta (approved):

PR #1976, opened 2022-08-16
ADCM-3074 `tests/base` to django tests

DanBalalan:


python/cm/tests/test_api.py
```diff
@@ -0,0 +1,832 @@
...
+from django.db import transaction
+from django.test import TestCase
+from django.urls import reverse
+from rest_framework.test import APIClient
+
+from init_db import init as init_adcm
+from rbac.upgrade.role import init_roles
+
+
+class TestBase(TestCase):
```
winmasta, line 30: There is base case test class in `python/adcm/tests/base.py`

python/cm/tests/test_api.py
```diff
@@ -0,0 +1,832 @@
...
+        with transaction.atomic():
+            response = self.client.post(
+                path=reverse("load-bundle"),
+                data={"bundle_file": bundle_name},
+            )
+        self.assertEqual(response.status_code, 200, msg=response.content)
+
+
+class TestAPI(TestBase):  # pylint: disable=too-many-public-methods
+    cluster = "adh42"
```
winmasta, line 67: Maybe it's better to use `setUp` instead of defining class propertioes

python/cm/tests/test_api.py
```diff
@@ -0,0 +1,832 @@
...
+    host = "test.host.net"
+    service = "ZOOKEEPER"
+    component = "ZOOKEEPER_SERVER"
+
+    def get_service_proto_id(self):
+        response = self.client.get(reverse("service-type"))
```
winmasta, line 73: Let's use one response variable name. `res` or `response`. Now there is plenty usages of `res`

python/cm/tests/test_api.py
```diff
@@ -0,0 +1,832 @@
...
+        self.assertEqual(response.status_code, 200, msg=response.content)
```
winmasta, line 74: It's better to use constantc from `rest_framework.statuses` for response codes. Assert message may help but in this case there can be attribute error instead of help

python/cm/tests/test_api.py
```diff
@@ -0,0 +1,832 @@
...
+        for service in response.json():
+            if service["name"] == self.service:
+                return service["id"]
+        return 0
```
winmasta, line 78: What does `0` means ?

winmasta (changes requested):
Need to discuss some general moments

python/cm/tests/test_api.py
DanBalalan: ¯\_(ツ)_/¯ Replaced with Runtime Error because this methods should return value

python/cm/tests/test_api.py
DanBalalan: done

python/cm/tests/test_api.py
DanBalalan: done

python/cm/tests/test_api.py
DanBalalan: done

python/cm/tests/test_api.py
DanBalalan: we decided to leave this testl like that

DanBalalan:
bender build

BenderArenadata:
Hey, @DanBalalan! Build has been launched!

a-alferov (approved):

winmasta (approved):

a-alferov:
bender build

BenderArenadata:

PR #1985, opened 2022-08-19
ADCM-3090 append audit logs to existing csvs

DanBalalan:


python/audit/management/commands/clearaudit.py
```diff
@@ -26,107 +26,107 @@
...
 
             cleared = self.__delete(target_operations, target_logins, target_objects)
 
-            self.__log('Finished.')
+            self.__log("Finished.")
             if cleared:
-                make_audit_log('audit', 'success', 'completed')
+                make_audit_log("audit", "success", "completed")
 
         except Exception as e:  # pylint: disable=broad-except
```
winmasta, line 92: It's better to minimize lines under `try` and use specific exeptions

python/audit/management/commands/clearaudit.py
```diff
@@ -140,11 +140,25 @@ def __prepare_csvs(self, *querysets, base_dir):
                 continue
 
             tmp_cvf_name = self.__get_csv_name(qs, now, base_dir)
-            with open(tmp_cvf_name, 'wt', newline='', encoding=self.encoding) as csv_file:
+            exists = os.path.exists(tmp_cvf_name)
```
winmasta, line 143: `pathlib` comes to replace `os.path` and `is_file` is better than `exists`

python/audit/management/commands/clearaudit.py
```diff
@@ -140,11 +140,25 @@ def __prepare_csvs(self, *querysets, base_dir):
...
+            mode = "at" if exists else "wt"
+            with open(tmp_cvf_name, mode, newline="", encoding=self.encoding) as csv_file:
                 writer = csv.writer(csv_file)
 
                 fields = [f.column for f in qs.model._meta.fields]
-                writer.writerow(fields)  # header
+                if not exists:
+                    writer.writerow(fields)  # header
+                else:
+                    with open(tmp_cvf_name, "rt", encoding=self.encoding) as csv_file:
```
winmasta, line 152: It's too complex to open file inside opened file context, maybe needs to be refactored

winmasta (changes requested):

python/audit/management/commands/clearaudit.py
DanBalalan: done

python/audit/management/commands/clearaudit.py
DanBalalan: refactored

python/audit/management/commands/clearaudit.py
DanBalalan: i think broad exception is needed to make correct auditlog

DanBalalan: partially fixed

python/audit/management/commands/clearaudit.py
```diff
@@ -139,37 +142,50 @@ def __prepare_csvs(self, *querysets, base_dir):
...
+        return header
 
     def __log(self, msg, method="info"):
         prefix = "Audit cleanup/archiving: "
         if method in ("exc", "exception"):
             log.warning("%sError in auditlog rotation", prefix)
             log.exception(msg)
         else:
-            msg = "Audit cleanup/archiving: " + str(msg)
+            msg = prefix + str(msg)
```
winmasta, line 189: We suupose to use f-interpolation for string concatenation and formatting

python/audit/management/commands/clearaudit.py
```diff
@@ -139,37 +142,50 @@ def __prepare_csvs(self, *querysets, base_dir):
...
+                base_dir,
+                f"audit_{now}_{self.archive_model_postfix_map[qs.model]}.csv",
+            )
+            header = self.__get_csv_header(tmp_cvf_name)
+            qs_fields = [f.column for f in qs.model._meta.fields]
+            if header:
+                if set(header) != set(qs_fields):
+                    self.__log(
+                        f"Fields of {qs.model._meta.object_name} was changed, "
+                        f"can\'t append to existing file. No archiving will be made",
```
winmasta, line 155: Backslash seems to be redundant

python/audit/management/commands/clearaudit.py
```diff
@@ -53,46 +54,48 @@ class Command(BaseCommand):
...
+        # get delete candidates
+        target_operations = AuditLog.objects.filter(operation_time__lt=threshold_date)
+        target_logins = AuditSession.objects.filter(login_time__lt=threshold_date)
+        objects_pk_to_delete = set()
+        for ao in AuditObject.objects.filter(is_deleted=True):
+            if not ao.auditlog_set.exclude(
+                pk__in=target_operations.values_list("pk", flat=True)
+            ).exists():
+                objects_pk_to_delete.add(ao.pk)
+        target_objects = AuditObject.objects.filter(pk__in=objects_pk_to_delete)
```
winmasta, line 80: We collect primary keys of instances of AuditObject, then we make queryset on AuditObject and filter them by collected primary keys. Why don't we just collect instances of AuditObject instead of collecting it's primary keys ?

winmasta (changes requested):

python/audit/management/commands/clearaudit.py
DanBalalan: We need to get all deleted AuditObjects for which there will be no AuditLog records after cleanup. So we need to check each audit_object is there any related AuditLog records that will not be deleted. As i know its impossible to construct such related-set-with-exclude lookup in django syntax

python/audit/management/commands/clearaudit.py
DanBalalan: fixed

python/audit/management/commands/clearaudit.py
DanBalalan: fixed

winmasta (approved):

python/audit/management/commands/clearaudit.py
DanBalalan: refactored

a-alferov (approved):

PR #1986, opened 2022-08-19
ADCM-3091 raise "400 INVALID_DATA" on policy update with no "id" key

DanBalalan:


SofiaTal (dismissed):

python/rbac/utils.py
```diff
@@ -15,11 +15,16 @@
...
+
 
 class BaseRelatedSerializer(serializers.Serializer):
     def to_internal_value(self, data):
         data = super().to_internal_value(data)
-        return data['id']
+        try:
+            return data["id"]
+        except KeyError as e:
+            raise AdcmEx("INVALID_DATA", f"Field {e} is required") from e
```
a-alferov, line 27:
```suggestion
        if "id" not in data:
            raise serializers.ValidationError("This field may not be empty.")
        return data.get("id")
```
`is required` invalid error for parent class

python/cm/errors.py
```diff
@@ -177,6 +177,11 @@
         rfs.HTTP_400_BAD_REQUEST,
         ERR,
     ),
+    "INVALID_DATA": (
```
a-alferov, line 180: it's too much

python/rbac/tests/tests_api/test_policy_unittests.py
```diff
@@ -0,0 +1,60 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from django.test import TestCase
+from django.urls import reverse
+from rest_framework.test import APIClient
```
a-alferov, line 15: please use Django client instead of DRF client

a-alferov (changes requested):

python/cm/errors.py
DanBalalan: removed

python/rbac/tests/tests_api/test_policy_unittests.py
DanBalalan: done

python/rbac/utils.py
DanBalalan: done

a-alferov (approved):

PR #2007, opened 2022-08-30
ADCM-3089 Add RBAC objects update audit object changes

winmasta:


python/audit/utils.py
```diff
@@ -86,6 +91,70 @@ def _get_deleted_obj(view: View, request: Request, kwargs) -> Model | None:
...
+    current_data = serializer_class(current_obj).data
+    current_fields = {k: v for k, v in current_data.items() if prev_data[k] != v}
+    if not current_fields:
+        return current_fields
+
+    object_changes = {
+        "current": current_fields,
+        "previous": {k: v for k, v in prev_data.items() if k in current_fields},
+    }
+    if object_changes["current"].get("password"):
```
a-alferov, line 117: `.get()` by default return `None`, if key is not found. Maybe use `if object_changes["current"].get("password") is not None:` ?

python/audit/utils.py
```diff
@@ -86,6 +91,70 @@ def _get_deleted_obj(view: View, request: Request, kwargs) -> Model | None:
...
+            serializer_class = UserAuditSerializer
+            model = User
+        elif view.__class__.__name__ == "PolicyViewSet":
+            serializer_class = PolicyAuditSerializer
+            model = Policy
+
+        if serializer_class:
+            current_obj = model.objects.filter(pk=view.kwargs["pk"]).first()
+            prev_data = serializer_class(model.objects.filter(pk=view.kwargs["pk"]).first()).data
+            if current_obj:
```
a-alferov, line 152: `current_obj` may be `ModelObject` or `None`. Maybe use `if current_obj is not None:` ?

a-alferov (approved):

PR #2011, opened 2022-08-30
ADCM-3115 Add user to audit log of finished actions

winmasta:


python/audit/utils.py
```diff
@@ -253,25 +252,26 @@ def audit_finish_task(obj, action_display_name: str, status: str) -> None:
...
+        object_name=obj.name,
         object_type=obj_type,
     )
     if status == "success":
         operation_result = AuditLogOperationResult.Success
     else:
         operation_result = AuditLogOperationResult.Fail
 
+    user = None
+    audit_log = AuditLog.objects.filter(audit_object=audit_object).first()
```
a-alferov, line 266: This request does not guarantee the determination of the correct user. We want the correct user to be the one who started the task.

a-alferov (changes requested):

python/audit/utils.py
winmasta: Refactored

python/audit/cases/common.py
```diff
@@ -83,17 +83,24 @@ def get_obj_name(obj: Model, obj_type: str) -> str:
...
+def get_or_create_audit_obj(
+        object_id: str,
+        object_name: str,
+        object_type: str,
+        action_id: int | None = None,
+) -> AuditObject:
     audit_object = AuditObject.objects.filter(
         object_id=object_id,
         object_type=object_type,
+        action_id=action_id,
```
a-alferov, line 95: Since `action_id` can be `null=True`, there is a chance to get 2 entries for one object

a-alferov (changes requested):

python/audit/cases/common.py
winmasta: refactored

a-alferov (approved):

PR #2015, opened 2022-09-01
ADCM-3064 Fix host fqdn validation, remove IntegrityError

winmasta:


a-alferov (changes requested):
Pls, add logic with host state according requirements

a-alferov (approved):

PR #2033, opened 2022-09-05
ADCM-3158 Rework user deactivating algorithm

SofiaTal:


conf/adcm/python_scripts/run_ldap_sync.py
```diff
@@ -142,7 +142,7 @@ def _sync_ldap_groups(self, ldap_groups: list) -> None:
             sys.stdout.write(f"Delete this group: {group}\n")
             group.delete()
         msg = "Sync of groups ended successfully."
-        msg += f"Couldn\'t synchronize groups: {error_names}\n" if error_names else ""
+        msg += f"Couldn't synchronize groups: {error_names}" if error_names else ""
```
winmasta, line 145:
```suggestion
        msg = f"{msg}Couldn't synchronize groups: {error_names}" if error_names else ""
```

a-alferov, line 145: I think it's better to write the whole line at once, and not use `+=`

winmasta (changes requested):

a-alferov (changes requested):

conf/adcm/python_scripts/run_ldap_sync.py
SofiaTal: Done as you suggest

winmasta (approved):

PR #2044, opened 2022-09-07
ADCM-3138 fix Internal Server error in case of request before deleting service

SofiaTal:


SofiaTal:
docker build

SofiaTal:
bender build

python/cm/adcm_config.py
```diff
@@ -517,12 +517,13 @@ def ui_config(obj, cl):  # pylint: disable=too-many-locals
...
     if obj.config:
-        cl = ConfigLog.objects.get(obj_ref=obj.config, id=obj.config.current)
-        obj_conf = cl.config
-    for c in conf:
-        if c.type != 'variant':
-            continue
-        c.limits['source']['value'] = cm.variant.get_variant(obj, obj_conf, c.limits)
+        cl = ConfigLog.objects.filter(obj_ref=obj.config, id=obj.config.current).first()
+        if cl:
+            obj_conf = cl.config
```
winmasta, line 522: Maybe it's redundant assign and we can use `cl.config` directly where it is necessary

SofiaTal: Sure! Delete obj_conf

PR #2056, opened 2022-09-12
ADCM-3153 add group_config config in inventory host_vars from all groups

SofiaTal:


python/cm/inventory.py
```diff
@@ -158,8 +159,18 @@ def get_provider_variables(provider: HostProvider, provider_config: dict = None)
     }
 
 
+def get_group_config(obj, host: Host) -> Optional[dict]:
```
winmasta, line 162:
```suggestion
def get_group_config(obj, host: Host) -> dict | None:
```

python/cm/unit_tests/utils.py
```diff
@@ -26,11 +26,11 @@ def gen_bundle(name='') -> models.Bundle:
     return models.Bundle.objects.create(**_gen_name(name or 'bundle_'), version='1.0.0')
 
 
-def gen_prototype(bundle: models.Bundle, proto_type) -> models.Prototype:
+def gen_prototype(bundle: models.Bundle, proto_type, name="") -> models.Prototype:
```
winmasta, line 29: Need type annotations for as more functions as possible

winmasta (changes requested):

winmasta (approved):

PR #2073, opened 2022-09-14
ADCM-3164 refactor existed tests in inventory, rework utils

SofiaTal:


python/cm/tests/utils.py
```diff
@@ -36,14 +38,14 @@
 )
 
 
-def _gen_name(prefix: str, name="name"):
+def gen_name(prefix: str):
     """Generate unique name"""
-    return {name: prefix + uuid4().hex}
+    return str(prefix + uuid4().hex)
```
winmasta, line 43: We use f-string interpolation if possible

winmasta (approved):

PR #2078, opened 2022-09-15
ADCM-3196 Rework logging audit

DanBalalan:


os/etc/service/wsgi/run
```diff
@@ -17,4 +17,4 @@ waitforinit
 
 echo "Run main wsgi application ..."
```
a-alferov, line 18: lacks:
```bash
exec 1>>"${adcmlog}/service_wsgi.out"
exec 2>>"${adcmlog}/service_wsgi.err"
```

a-alferov (changes requested):

os/etc/service/wsgi/run
DanBalalan: added

a-alferov (approved):

PR #2114, opened 2022-09-28
ADCM-3222 fix 500 While launching action with activatable config

DanBalalan:


python/cm/job.py
```diff
@@ -191,7 +191,7 @@ def check_action_config(
     if obj.config:
         cl = ConfigLog.objects.get(obj_ref=obj.config, id=obj.config.current)
         obj_conf = cl.config
-        current_attr = cl.attr
+        current_attr = cl.attr or None  # convert {} (empty dict) to None
```
a-alferov, line 194: What for ?

a-alferov (changes requested):

python/cm/job.py
DanBalalan: to prevent KeyError in cm.adcm_config.check_attr when comparing attr's kaeys and current_attr's (cluster attr == {}) keys

kuhella:
bender build

BenderArenadata:
Hey, @kuhella! Build has been launched!

python/cm/job.py
```diff
@@ -187,12 +187,11 @@ def check_action_config(
         return {}, {}
     if not conf:
         err("TASK_ERROR", "action config is required")
-    obj_conf, current_attr = None, None
+    obj_conf = None
     if obj.config:
-        cl = ConfigLog.objects.get(obj_ref=obj.config, id=obj.config.current)
-        obj_conf = cl.config
-        current_attr = cl.attr
```
a-alferov, line 194: It is necessary to completely remove this block, and pass the configuration of the action to the function `variant.process_variant(obj, spec, conf)`

DanBalalan: done

a-alferov (approved):

PR #2115, opened 2022-09-28
Test on known bug with activatable group in action's config

Sealwing:


Sealwing:
Need to add case https://tracker.yandex.ru/ADCM-3222

chudasovv (approved):

PR #2117, opened 2022-09-29
ADCM-3235 remove tmp dir from archiving, fix naming rules, process deleted sub_action

SofiaTal:


python/adcm/utils.py
```diff
@@ -25,8 +25,10 @@ def get_obj_type(obj_type: str) -> str:
 
 
 def str_remove_non_alnum(value: str) -> str:
-    return "".join(ch for ch in value if ch.isalnum())
-
+    result = "".join(ch.lower() for ch in value if (ch.isalnum() or ch == " ")).replace(" ", "-")
```
winmasta, line 28:
```suggestion
    result = "".join(ch.lower().replace(" ", "-") for ch in value if (ch.isalnum() or ch == " "))
```

python/api/job/views.py
```diff
@@ -124,31 +124,40 @@ def get_task_download_archive_file_handler(task: TaskLog) -> io.BytesIO:
...
+                        value=job.sub_action.display_name
+                    ) or str_remove_non_alnum(value=job.sub_action.name)
+            else:
+                dir_name_suffix = task_dir_name_suffix
 
             directory = Path(settings.RUN_DIR, str(job.pk))
             if directory.is_dir():
-                for log_file in Path(settings.RUN_DIR, str(job.pk)).iterdir():
-                    tarinfo = tarfile.TarInfo(f"{job.pk}-{dir_name_suffix}/{log_file.name}")
+                for log_file in [
```
a-alferov, line 147: Please make a separate variable, this code smells bad

python/api/job/views.py
```diff
@@ -124,31 +124,40 @@ def get_task_download_archive_file_handler(task: TaskLog) -> io.BytesIO:
...
+                    item for item in Path(settings.RUN_DIR, str(job.pk)).iterdir() if item.is_file()
+                ]:
+                    tarinfo = tarfile.TarInfo(
+                        f"{job.pk}-{dir_name_suffix}".strip("-") + f"/{log_file.name}"
```
a-alferov, line 151: Use pls f-string, do not use `+`

a-alferov (changes requested):

python/api/job/views.py
SofiaTal: I need to use strip() to delete extra "-" if dir_name_suffix is absent (It is possible if subaction is None and bundle deleted). I can't strip the whole string and alternate variant is ternar operator and checking dir_name_suffix. I think way with "+" better and shorter.

a-alferov: You can use:
```suggestion
                        f'{f"{job.pk}-{dir_name_suffix}".strip("-")}/{log_file.name}'
```
or:
```suggestion
                        str(Path(f"{job.pk}-{dir_name_suffix}".strip("-"), f"{log_file.name}"))
```

a-alferov (changes requested):
Add pls `download` field in TaskLog serializer

python/api/job/views.py
SofiaTal: Done

a-alferov (dismissed):

a-alferov (approved):

PR #2122, opened 2022-09-30
ADCM-3065 Maintenance Mode for  components/services

SofiaTal:


python/cm/migrations/0095_increase_max_length.py
```diff
@@ -17,7 +18,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("cm", "0093_auto_20220928_0556"),
+        ("cm", "0094_increase_of_maxlength_to_postgress"),
```
winmasta, line 21: All migrations could be merged into one

PR #2133, opened 2022-10-05
ADCM-3257 fix naming of issue in case of cluster or host renaming

SofiaTal:


python/cm/issue.py
```diff
@@ -297,17 +297,24 @@ def _gen_issue_name(obj: ADCMEntity, cause: ConcernCause) -> str:
     return f"{obj} has issue with {cause.value}"
 
 
+def _create_concern_item(obj: ADCMEntity, issue_cause: ConcernCause):
```
a-alferov, line 300:
```suggestion
def _create_concern_item(obj: ADCMEntity, issue_cause: ConcernCause) -> ConcernItem:
```

a-alferov (changes requested):

PR #2136, opened 2022-10-06
ADCM-3263 rename disabling_cause, change messages from disabling cause, show actions for host in mm

SofiaTal:


python/api/action/views.py
```diff
@@ -193,24 +190,12 @@ def check_action_perm(self, action, obj):
             raise PermissionDenied()
 
     @staticmethod
-    def check_disabling_cause(action, obj):
-        if isinstance(obj, Host) and obj.maintenance_mode:
+    def check_start_impossible_reason(action, obj):
+        set_start_impossible_reason(obj, action)
```
winmasta, line 194: Method with "check" in name shouldn't "set" anything

python/api/action/views.py
```diff
@@ -193,24 +190,12 @@ def check_action_perm(self, action, obj):
             raise PermissionDenied()
 
     @staticmethod
-    def check_disabling_cause(action, obj):
-        if isinstance(obj, Host) and obj.maintenance_mode:
+    def check_start_impossible_reason(action, obj):
```
winmasta, line 193: Type annotations can be added

python/api/action/views.py
```diff
@@ -163,7 +160,7 @@ def get(self, request, *args, **kwargs):
             self.get_queryset(),
             id=action_id,
         )
-        set_disabling_cause(obj, action)
+        set_start_impossible_reason(obj, action)
```
winmasta, line 163: Passing parameters only as named agreed. Setter of Action class can be defined instead of separate function

python/api/utils.py
```diff
@@ -37,6 +37,13 @@
 )
 from cm.upgrade import get_upgrade
 
+REASON_MSG_DICT = {
```
winmasta, line 40: Should we really define some list of possible reasons ? May be we can just use some arbitrary string, save it in DB, and pass it for API request

winmasta (approved):

PR #2149, opened 2022-12-27
Fix secret key location and ensure existence

giggsoff:
Two fixes spitted into separate commits:

1. generate_secret_key [stores](https://github.com/MickaelBergem/django-generate-secret-key/blob/91d974a1870a40bdaac93f7037a6a0af85eeaaa6/django_generate_secret_key/management/commands/generate_secret_key.py#L40) the key into the current location, so we should use it from there.
2. It is nice to have the secret file in place in order to not invalidate sessions (for example download logs of jobs) after restart of container (as we will [generate](https://github.com/arenadata/adcm/blob/d41b7685535b060fbdc7251c6a6b0e127df1b862/python/adcm/settings.py#L69) new key on fly if no file stored), so let's check for the key existence to generate it.

### DIFF
--- a/os/etc/sv/init/run
+++ b/os/etc/sv/init/run
@@ -44,7 +44,8 @@ key=/root/.ssh/id_rsa
 if [ ! -f "${adcmvar}"/cluster.db ]; then
     echo "Assume first run, generate Django secret key"
     "${adcmroot}"/python/manage.py generate_secret_key
-    mv "${adcmroot}/python/secretkey.txt" "${adcmsecretfile}"
+    # generate_secret_key creates file in the current directory with secretkey.txt name
+    mv secretkey.txt "${adcmsecretfile}"
 fi
 
 if [ ! \( -f "${adcmuserconf}/ssl/cert.pem" -a -f "${adcmuserconf}/ssl/key.pem" \) ]; then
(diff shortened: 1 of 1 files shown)


a-alferov:
@giggsoff Thanks for participating. We fixed this in the latest release. https://github.com/arenadata/adcm/blob/2022.12.26.18/os/etc/sv/init/run#L47
 In the future, we will completely get rid of this file.

giggsoff:
> @giggsoff Thanks for participating. We fixed this in the latest release. https://github.com/arenadata/adcm/blob/2022.12.26.18/os/etc/sv/init/run#L47 In the future, we will completely get rid of this file.

The problem in the file location. You point me onto the code with `mv "${adcmroot}/python/secretkey.txt" "${adcmsecretfile}"`, but it should be `mv "secretkey.txt" "${adcmsecretfile}"` (see my comment in the PR).

a-alferov:
@giggsoff Yes you are right. Now it still works thanks to this code: https://github.com/arenadata/adcm/blob/master/python/adcm/settings.py#L52 
But in the next release, we will completely get rid of storing the secret key on the file system https://github.com/arenadata/adcm/commit/d57ab68e11ead55f5295e6f2b87473c8a8768dff

